PR #3, opened 2019-07-29
Pull in relevant parts of type erasure experiment

kyleve:
Type erasure wrappers (which allowed nice types to flow up to the table view for row elements) seemed nice in theory, but they made the API pretty wordy... Lots of `Element` generics absolutely everywhere, which wasn't great. Instead, continue allowing heterogeneous content in the table, but add helper methods to compare elements for testing purposes.


Sources/Binding.swift
```diff
@@ -49,42 +64,46 @@ public final class Binding<Element>
...
     {
         switch self.state {
-        case .initializing, .idle, .discarded: break
-        
-        case .updating(let context, let update, _):
-            self.state = .idle(context, update)
+        case .initializing, .updating, .discarded: break
+            
+        case .new(let new):
+            new.context.bindAny(to: self)
```
kyleve: Need to set to `.updating` as well.

PR #9, opened 2019-08-05
Updated state management APIs

kyleve:
Update state management APIs by moving them into the table view, instead of having a wrapping presenter.


Sources/TableView/TableView.swift
```diff
@@ -53,6 +53,38 @@ public final class TableView : UIView
         self.updateVisibleSlice(for: .contentChanged(animated: animated))
     }
     
+    // MARK: Content Sources
+    
+    private var sourcePresenter : TableViewSourcePresenter? = nil
+    
+    @discardableResult
+    public func setSource<Source:TableViewSource>(initial : Source.Input, source : Source, animated : Bool = false) -> ValueAccess<Source.Input>
```
kyleve: Consider: run setContent through this method so everything follows one path, and setting content kills the sourcePresenter so we dont end up in an inconsistent state.

kyleve:
Ended up keeping the tableview-based source – instead made the content be backed by the source presenter.

Sources/TableView/TableViewSource.swift
```diff
@@ -0,0 +1,155 @@
...
+    private let set : (Value) -> ()
+    
+    internal init(get : @escaping () -> Value, set : @escaping (Value) -> ())
+    {
+        self.get = get
+        self.set = set
+    }
+}
+
+public final class State<Value:Equatable>
```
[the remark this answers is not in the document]
kyleve: `SourceState`

Sources/TableView/TableView.swift
kyleve: Done

PR #13, opened 2019-08-08
Split into core, table, and collection podspecs

kyleve:
This allows us to use terser names like "Row" and "Section" as top level items in each pod – since its rare you'll need both in a single file.


kyleve:
TODO: The way CI is configured now isnt great since it builds the whole demo app... Technically fine, but may want to shave down and exclude blueprint? Or no?

kyleve:
TODO: Keep ListableCore public, with the components that are needed publicly. Make a new ListableInternal pod, which will have tests, but its sources will be directly included in the ListableTableView and ListableCollectionView pods. This allows keeping internal internal.

PR #21, opened 2019-09-23
Move to collection view-backed, UIView only API

kyleve:
This PR re-implements most of Listable on top of UICollectionView (though, hidden from the public API). This allows much more substantial customization of appearance of the list view, which is not possible with UITableView. I also change the public APIs to deal purely in views, not in cells.


Listable/Sources/ListView.swift
```diff
@@ -0,0 +1,692 @@
...
+    //
+    // MARK: Setting & Getting Content
+    //
+    
+    public var content : Content {
+        get { return self.storage.allContent }
+        set { self.set(content: false, newValue) }
+    }
+    
+    public func set(content animated : Bool = false, _ block : ContentBuilder.Build)
```
kyleve: Don't love how these are named... Should probably go back to `public func setContent(animated : Bool = false, _ content : Content)`

Listable/Sources/ListView.swift
```diff
@@ -0,0 +1,692 @@
...
+        case .scrolledToTop:
+            self.updatePresentationStateWith(firstVisibleIndexPath: IndexPath(item: 0, section: 0), for: reason)
+        }
+    }
+    
+    private func updatePresentationStateWith(firstVisibleIndexPath indexPath: IndexPath?, for reason : Content.Slice.UpdateReason)
+    {
+        let indexPath = indexPath ?? IndexPath(item: 0, section: 0)
+        
+        let slice = self.content.sliceTo(indexPath: indexPath, plus: Content.Slice.defaultSize)
```
kyleve: self.storage.allContent

Listable/Sources/Sizing.swift
```diff
@@ -0,0 +1,107 @@
+//
+//  Height.swift
```
kyleve: Sizing.swift

PR #22, opened 2019-09-26
Reset version numbers to real version numbers so we can begin bumping when consumed by SPOS

kyleve:


kyleve:
@bizzemfrog FYI landing this

PR #24, opened 2019-09-26
Fix HeaderFooterViewElement implementation of apply

watt:


kyleve:
Mind bumping the pod version as well?

PR #25, opened 2019-10-30
Separate callbacks for display and end display

kyleve:


kyleve:
@bizzemfrog fyi based on seeing this used in transactions

PR #30, opened 2019-11-01
Frame & Blueprint Fixes

kyleve:
- Do not display content until the table view has a size, and properly init with a frame from blueprint.

### DIFF
--- a/BlueprintLists/Sources/List.swift
+++ b/BlueprintLists/Sources/List.swift
@@ -39,7 +39,7 @@ public struct List : BlueprintUI.Element
     {
         return ListView.describe { config in
             config.builder = {
-                return ListView(frame: .zero, appearance: self.appearance)
+                return ListView(frame: bounds, appearance: self.appearance)
             }
             
             config.apply { listView in
--- a/Listable/Sources/ListView.swift
+++ b/Listable/Sources/ListView.swift
@@ -141,6 +141,26 @@ public final class ListView : UIView
     
     // MARK: UIView
     
+    public override var frame: CGRect {
+        didSet {
+            /**
+             Once the view actually has a size, we can provide content.
+            
+             There's no value in having content with no view size, as we cannot
+             size cells otherwise.
+             */
+            
+            let fromEmpty = oldValue.isEmpty && self.bounds.isEmpty == false
+            let toEmpty = oldValue.isEmpty == false && self.bounds.isEmpty
+            
+            if fromEmpty {
+                self.updatePresentationState(for: .transitionedToBounds(isEmpty: false))
+            } else if toEmpty {
+                self.updatePresentationState(for: .transitionedToBounds(isEmpty: true))
+            }
+        }
+    }
+    
     public override var backgroundColor: UIColor? {
         didSet {
             self.collectionView.backgroundColor = self.backgroundColor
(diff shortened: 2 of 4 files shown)


BlueprintLists/Sources/List.swift
```diff
@@ -39,7 +39,7 @@ public struct List : BlueprintUI.Element
     {
         return ListView.describe { config in
             config.builder = {
-                return ListView(frame: .zero, appearance: self.appearance)
+                return ListView(frame: bounds, appearance: self.appearance)
```
kyleve: Doh...

Listable/Sources/ListView.swift
```diff
@@ -223,6 +231,49 @@ public final class ListView : UIView
...
+        
+        static func == (lhs : VisibleItem, rhs : VisibleItem) -> Bool
+        {
+            return lhs.value === rhs.value
+        }
+    }
+    
+    private var visibleItems : Set<VisibleItem> = Set()
+    
+    private func updateDisplayedItemStates()
```
kyleve: This name sucks... `updateDisplayedItems`?

Listable/Sources/PresentationState.swift
```diff
@@ -568,8 +588,6 @@ final internal class PresentationState
             let cell = (anyCell as! ItemElementCell<Element>)
             
             self.visibleCell = cell
-            
-            self.model.onDisplay?(self.model.element)
```
kyleve: Since we reload the table, do diffs, etc, this got called far too frequently.. Instead now base this on the table view's visible indexes within `updateDisplayedItemStates`

Listable/Sources/ListView.swift
```diff
@@ -337,10 +392,14 @@ public final class ListView : UIView
...
         if animated {
-            view.performBatchUpdates(batchUpdates)
+            view.performBatchUpdates(batchUpdates) { _ in
+                self.updateDisplayedItemStates()
+            }
         } else {
             UIView.performWithoutAnimation {
-                view.performBatchUpdates(batchUpdates)
+                view.performBatchUpdates(batchUpdates) { _ in
+                    self.updateDisplayedItemStates()
```
kyleve: This needs to move a layer up to `updatePresentationStateWith` – since this method is only called for diffs.

BlueprintLists/Sources/List.swift
bizzemfrog: Doh indeed

Listable/Sources/ListView.swift
```diff
@@ -250,14 +308,18 @@ public final class ListView : UIView
...
+            self.updatePresentationStateWith(firstVisibleIndexPath: indexPath, for: reason)
         }
     }
     
     private func updatePresentationStateWith(firstVisibleIndexPath indexPath: IndexPath?, for reason : Content.Slice.UpdateReason)
     {
         let indexPath = indexPath ?? IndexPath(item: 0, section: 0)
         
-        let slice = self.storage.allContent.sliceTo(indexPath: indexPath, plus: Content.Slice.defaultSize)
+        let slice = self.frame.isEmpty ? Content.Slice() : self.storage.allContent.sliceTo(indexPath: indexPath, plus: Content.Slice.defaultSize)
```
bizzemfrog: is frame is empty why not just early return here?

Listable/Sources/ListView.swift
```diff
@@ -268,11 +330,14 @@ public final class ListView : UIView
             self.performBatchUpdates(
                 with: diff,
                 animated: reason.animated,
-                onBeginUpdates: updateData
-            )
+                onBeginUpdates: updateData,
+                completion: { _ in
+                    self.updateVisibleItems()
+            })
```
bizzemfrog: nit: put the close paren at the same indentation as `completion: {`

Listable/Sources/ListView.swift
kyleve: In theory if there's already content (eg we went from a zero to non zero to zero frame), we'd still want to remove things – but I also need to implement that (oops – forgot about the non-zero to zero path).

Listable/Sources/ListView.swift
kyleve: I _think_ this is what Xcode just does, but will fix if it isn't

kyleve: Confirmed this is how Xcode does it, so just leaving so I'm not fighting its formatting.

Listable/Sources/ListView.swift
kyleve: Fixed

Listable/Sources/ListView.swift
kyleve: Fixed

Listable/Sources/ListView.swift
kyleve: Fixed

PR #34, opened 2019-11-04
[WIP] Start implementing Snapshot tests, and implement scrollTo so we can test async snapshot tests

kyleve:


kyleve:
Closing this for now. Wanna figure out a better shape for snapshots where they don't break for subpixel reasons.

PR #37, opened 2019-11-05
Update all cells on screen whenever an update happens

kyleve:


Listable/Sources/Content.swift
```diff
@@ -153,23 +153,6 @@ internal extension Content
             
             case transitionedToBounds(isEmpty : Bool)
             
-            var diffsChanges : Bool {
```
kyleve: No longer need this distinction since CollectionView doesnt have artifacts during non-updated diffs.

Listable/Sources/Internal/Diff.swift
```diff
@@ -775,7 +775,7 @@ extension SectionedDiff.ItemChanges
         
         self.noChange.forEach {
             var value = new[$0.newIndex.item]
-            updated($0.oldValue, $0.newValue, &value)
+            noChange($0.oldValue, $0.newValue, &value)
```
kyleve: Oops...

Listable/Sources/ListView.swift
```diff
@@ -300,10 +327,14 @@ public final class ListView : UIView
             self.updatePresentationStateWith(firstVisibleIndexPath: indexPath, for: reason)
             
         case .didEndDecelerating:
-            self.updatePresentationStateWith(firstVisibleIndexPath: indexPath, for: reason)
+            if presentationStateTruncated {
```
kyleve: Performance optimization

Listable/Sources/ListView.swift
```diff
@@ -374,30 +399,17 @@ public final class ListView : UIView
...
-                section.header?.applyToVisibleView()
-                section.footer?.applyToVisibleView()
-            }
-            
-            diff.changes.noChange.forEach { change in
-                let section = self.storage.presentationState.sections[change.newIndex]
-                
-                section.header?.applyToVisibleView()
-                section.footer?.applyToVisibleView()
+            self.visibleSections.forEach {
```
kyleve: This and below are the actual change.

Listable/Sources/Content.swift
kyleve: Buuut it turns out this is hella slow now, bringing it back...

PR #41, opened 2019-11-06
Keyboard Support

kyleve:
Support updating content inset when the table changes.

### DIFF
--- a/Listable/Sources/Internal/KeyboardObserver.swift
+++ b/Listable/Sources/Internal/KeyboardObserver.swift
@@ -1,0 +1,158 @@
+//
+//  KeyboardObserver.swift
+//  Listable
+//
+//  Created by Kyle Van Essen on 11/5/19.
+//
+
+import Foundation
+
+
+protocol KeyboardObserverDelegate : AnyObject
+{
+    func keyboardFrameWillChange(observer : KeyboardObserver)
+}
+
+/**
+ iOS Docs: https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/KeyboardManagement/KeyboardManagement.html
+ */
+final class KeyboardObserver
+{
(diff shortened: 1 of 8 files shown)


Listable/Sources/ListView.swift
```diff
@@ -224,6 +232,24 @@ public final class ListView : UIView
         }
     }
     
+    public override func willMove(toWindow newWindow: UIWindow?)
```
kyleve: This needs to be in didMove, not willMove.

Listable/Sources/ListView.swift
```diff
@@ -224,6 +232,24 @@ public final class ListView : UIView
...
+    {
+        super.willMove(toWindow: newWindow)
+        
+        if newWindow != nil {
+            self.setContentInsetWithKeyboardFrame()
+        }
+    }
+    
+    public override func willMove(toSuperview newSuperview: UIView?)
```
kyleve: This needs to be in didMove, not willMove.

Listable/Sources/Internal/KeyboardObserver.swift
```diff
@@ -0,0 +1,158 @@
...
+{
+    private let center : NotificationCenter
+    
+    weak var delegate : KeyboardObserverDelegate?
+    
+    init(center : NotificationCenter = .default)
+    {
+        self.center = center
+
+        self.center.addObserver(self, selector: #selector(keyboardFrameChanged(_:)), name: UIWindow.keyboardWillChangeFrameNotification, object: nil)
```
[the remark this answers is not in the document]
kyleve: Nevermind, `If your app targets iOS 9.0 and later or macOS 10.11 and later, you don't need to unregister an observer in its dealloc method. Otherwise, you should call removeObserver(_:name:object:) before observer or any object passed to this method is deallocated.`

PR #44, opened 2019-11-08
Add contentIdentifier to allow for wholesale reloading of the table without animation state

kyleve:


Listable/Sources/Internal/PresentationState.swift
```diff
@@ -86,6 +86,8 @@ final class PresentationState
         
     private(set) var containsAllItems : Bool
     
+    private(set) var contentIdentifier : AnyHashable?
```
kyleve: Dont need this anymore

Listable/Sources/Content.swift
```diff
@@ -213,7 +218,7 @@ internal extension Content
                 case .didEndDecelerating: return false
                 case .scrolledToTop: return false
                     
-                case .contentChanged(let animated): return animated
+                case .contentChanged(let animated, let identifierChanged): return animated && identifierChanged == false
```
kyleve: Might consider bringing back diffsChanges for this...

PR #63, opened 2019-11-20
Add helpers to sizing so we don't always need to do .sizeThatFits(.noConstraint).

kyleve:


Listable/Sources/Content.swift
```diff
@@ -199,7 +199,7 @@ internal extension Content
 {
     struct Slice
     {
-        static let defaultSize : Int = 250
+        static let defaultSize : Int = 100_000
```
kyleve: Don't merge this!

PR #82, opened 2019-12-12
Support overscroll footers

kyleve:
This PR adds support for "overscroll" footers, aka what is showed when a user drags up so that content below the end of the table is visible.

It also cleans up the layout code for stickyheaders to be similar to the code for overscroll – it's easier to do this on each layout prepare() call (and usually needed), so avoid lumping this into the NeededLayoutType enum.

No scroll:

Screen Shot 2019-12-11 at 9 14 10 PM

Scrolled:

Screen Shot 2019-12-11 at 9 14 25 PM

### DIFF
--- a/Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
+++ b/Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
@@ -18,6 +18,8 @@ extension ListViewLayout
         case sectionHeader = "Listable.ListViewLayout.SectionHeader"
         case sectionFooter = "Listable.ListViewLayout.SectionFooter"
         
+        case overscrollFooter = "Listable.ListViewLayout.OverscrollFooter"
+        
         var zIndex : Int {
             switch self {
             case .listHeader: return 1
@@ -25,6 +27,8 @@ extension ListViewLayout
                 
             case .sectionHeader: return 2
             case .sectionFooter: return 1
+                
+            case .overscrollFooter: return 1
             }
         }
         
@@ -36,6 +40,8 @@ extension ListViewLayout
                 
             case .sectionHeader: return IndexPath(item: 0, section: section)
             case .sectionFooter: return IndexPath(item: 0, section: section)
+                
+            case .overscrollFooter: return IndexPath(item: 0, section: 0)
             }
         }
     }
@@ -58,6 +64,8 @@ internal extension ListViewLayout
         let header : SupplementaryItemLayoutInfo?
         let footer : SupplementaryItemLayoutInfo?
         
+        let overscrollFooter : SupplementaryItemLayoutInfo?
+        
         private(set) var sections : [SectionLayoutInfo]
         
         //
@@ -73,6 +81,7 @@ internal extension ListViewLayout
             
             self.header = nil
             self.footer = nil
+            self.overscrollFooter = nil
             
             self.sections = []
         }
@@ -107,6 +116,16 @@ internal extension ListViewLayout
                     kind: SupplementaryKind.listFooter,
                     direction: appearance.direction,
                     layout: delegate.layoutForListFooter(in: collectionView)
+                )
+            }()
+            
+            self.overscrollFooter = {
+                guard delegate.hasOverscrollFooter(in: collectionView) else { return nil }
+                
+                return SupplementaryItemLayoutInfo(
+                    kind: SupplementaryKind.overscrollFooter,
+                    direction: appearance.direction,
+                    layout: delegate.layoutForOverscrollFooter(in: collectionView)
                 )
             }()
             
@@ -160,7 +179,7 @@ internal extension ListViewLayout
             
             if let header = self.header {
                 if rect.intersects(header.visibleFrame) {
-                    attributes.append(header.layoutAttributes(with: SupplementaryKind.listHeader.indexPath(in: 0)))
+                    attributes.append(header.layoutAttributes(with: header.kind.indexPath(in: 0)))
                 }
             }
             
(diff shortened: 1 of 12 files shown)


Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
```diff
@@ -278,6 +305,10 @@ internal extension ListViewLayout
             
             let direction = self.appearance.direction
             
+            // TODO: Using lst_safeAreaInsets is likely wrong if this is called before the safe area is applied by the view hierarchy.
```
kyleve: Found this while working on the overscroll stuff. Going to fix in a separate PR.

Listable/Sources/ListView/ListView.Delegate.swift
```diff
@@ -220,6 +223,30 @@ extension ListView
...
+        }
+        
+        func heightForOverscrollFooter(in collectionView : UICollectionView, width : CGFloat, layoutDirection : LayoutDirection) -> CGFloat
+        {
+            let footer = self.presentationState.overscrollFooter!
+            
+            return footer.height(
+                width: width,
+                layoutDirection: layoutDirection,
+                defaultHeight: self.view.layout.appearance.sizing.listFooterHeight,
```
kyleve: Add an `overscrollFooterHeight`to Sizing.

Listable/Sources/ListView/ListView.Delegate.swift
```diff
@@ -220,6 +223,30 @@ extension ListView
...
+                measurementCache: self.headerMeasurementCache
```
kyleve: footerMeasurementCache. Also, can those just be the same cache?

kyleve: Turns out they can, footerCache was never used. Fixing.

Listable/Sources/ListView/ListView.Delegate.swift
kyleve: Done

Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
kyleve: Never mind, gonna fix since it makes other stuff cleaner.

Listable/Sources/Internal/Layout/ListViewLayout.swift
```diff
@@ -171,15 +175,6 @@ class ListViewLayout : UICollectionViewLayout
         self.neededLayoutType.merge(with: context)
     }
     
-    override func invalidationContext(forBoundsChange newBounds: CGRect) -> UICollectionViewLayoutInvalidationContext
-    {
-        let context = super.invalidationContext(forBoundsChange: newBounds) as! InvalidationContext
-        
-        context.updateHeaders = self.appearance.layout.stickySectionHeaders
```
kyleve: Removed this property; we need to update this every time alongside the overscroll values, so its not useful to keep.

Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
```diff
@@ -305,6 +332,34 @@ internal extension ListViewLayout
...
+            let direction = self.appearance.direction
+            
+            let contentHeight = direction.height(for: self.contentSize)
+            let viewHeight = direction.height(for: collectionView.bounds.size)
+            
+            // Overscroll positioning is done after we've sized the layout, because the overscroll footer does not actually
+            // affect any form of layout or sizing. It appears only once the scroll view has been scrolled outside of its normal bounds.
+            
+            if contentHeight >= viewHeight {
+                footer.y = contentHeight + direction.bottom(with: collectionView.lst_safeAreaInsets)
```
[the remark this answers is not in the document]
kyleve: Nope, bottom is right here.

Listable/Sources/ListView/ListView.Delegate.swift
```diff
@@ -220,6 +224,30 @@ extension ListView
             return footer.anyModel.layout
         }
         
+        func hasOverscrollFooter(in collectionView : UICollectionView) -> Bool
+        {
+            return self.presentationState.overscrollFooter != nil
+        }
+        
+        func heightForOverscrollFooter(in collectionView : UICollectionView, width : CGFloat, layoutDirection : LayoutDirection) -> CGFloat
```
bizzemfrog: nit: each arg on its own new line

Listable/Sources/ListView/ListView.Delegate.swift
```diff
@@ -220,6 +224,30 @@ extension ListView
...
+        {
+            let footer = self.presentationState.overscrollFooter!
```
bizzemfrog: might want to `guard let` here, and and return 0 in the else block.

Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
kyleve: This likely also needs contentInset.

kyleve: (For keyboards)

kyleve: Never mind, this works fine.

Listable/Sources/ListView/ListView.Delegate.swift
kyleve: The way this API works right now is similar to the UICollectionViewDataSource's `collectionView:viewForSupplementaryElementOfKind:atIndexPath:`, which if called, requires you to return a view – so I figured doing a force unwrap would be fine – since it'll lead to undefined behaviour anyway.

Listable/Sources/ListView/ListView.Delegate.swift
kyleve: Done, and also for the other methods in this file, and the DataSource file.

PR #83, opened 2019-12-14
Additional overscroll fixes

kyleve:
- Support for horizontal layouts.
- Fix missing use of contentInset for longer lists.

### DIFF
--- a/Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
+++ b/Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
@@ -172,6 +172,8 @@ internal extension ListViewLayout
         //
         // MARK: Fetching Elements
         //
+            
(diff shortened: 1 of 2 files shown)


Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
```diff
@@ -352,7 +355,7 @@ internal extension ListViewLayout
             // affect any form of layout or sizing. It appears only once the scroll view has been scrolled outside of its normal bounds.
             
             if contentHeight >= viewHeight {
-                footer.y = contentHeight + direction.bottom(with: collectionView.lst_safeAreaInsets)
+                footer.y = contentHeight + direction.bottom(with: collectionView.contentInset) + direction.bottom(with: collectionView.lst_safeAreaInsets)
```
kyleve: This does break keyboards though... May only want to do this for horizontal stuff...

PR #90, opened 2020-01-11
Ensure we always refresh and update supplementary views when they change

kyleve:
Fixes https://jira.sqcorp.co/browse/UI-489

## Notes
Important changes are within `Listable/Sources/**`

## The Problem

Within collection views, supplementary views (how you model headers, footer) are attached to individual index paths. So, Listable models headers and footers as attached to (0,0) for list headers and footers, and (sectionIndex, 0) for section headers and footers. All good so far.

The problem arises when you want to swap out a header or footer without changing the row at the (x, 0) index path. The collection view does not know to re-query for those associated headers, because we didn't actually reload or change that (x, 0) item.

## The Solution (First Try)

At first I thought: Ok, if we need to reload the item at (1,0) to reload the headers and footers in section 1, let's do that if they changed, but the item didn't. Unfortunately this gets brittle around insertions of rows and whole sections. Eg, I can only reload the item if it (a) wasn't already reloaded this cycle, **and** (b), it wasn't inserted this cycle, **and** (c) if the whole section its in wasnt reloaded this cycle. Pretty complicated to track all this, so I avoided this.

## The Actual Solution
Where I ended up at, and is implemented here, is we _always_ return supplementary views to the collection view (`SupplementaryContainerView`) for each header and footer, but give it a zero height when there is no backing view from the developer. Then, when the developer puts in a header or footer, we add it to the `SupplementaryContainerView` as a content view. When they remove a header or footer, we remove the content view, and the height of the supplementary view again becomes 0. This means we can fully control when we add or remove these supplementary views, without having to trick the collection view into reloading items at random times.

## Remaining Work
The animations when swapping out the headers and footers don't exist yet. But, want to get this landed so it stops crashing in the SPOS transactions applet.


Listable/Sources/Internal/Assertions.swift
```diff
@@ -0,0 +1,39 @@
...
+//  Assertions.swift
+//  Listable
+//
+//  Created by Kyle Van Essen on 1/10/20.
+//
+
+import Foundation
+
+
+public func listableFatal(_ message: @autoclosure () -> String = String(), file: StaticString = #file, line: UInt = #line) -> Never
```
kyleve: Public because we also need this from BlueprintLists.

Listable/Tests/ListView/ListViewTests.swift
```diff
@@ -7,7 +7,94 @@
...
 
+@testable import Listable
+
+
+
 class ListViewTests: XCTestCase
 {
+    func test_supplementary_views_without_content()
+    {
+        // TODO...
```
kyleve: Need to finish this.

Listable/Tests/ListView/ListViewTests.swift
```diff
@@ -7,7 +7,94 @@
...
+        listView.setContent { list in
+            list.animatesChanges = false
+            
+            list += Section(identifier: "a-section")
+            list.content.overscrollFooter = HeaderFooter(with: TestSupplementary(), appearance: TestSupplementary.Appearance())
+        }
+        
+        listView.collectionView.contentOffset.y = 100
+        listView.collectionView.layoutIfNeeded()
+        self.waitForOneRunloop()
```
kyleve: Remove these?

kyleve: Nope, needed to make each layout pass run correctly. Even layoutIfNeeded doesn't to the right thing.

Listable/Sources/Internal/Layout/ListViewLayout.LayoutInfo.swift
```diff
@@ -409,15 +427,19 @@ internal extension ListViewLayout
             // Header
             //
             
-            if let header = self.header {
+            performLayout(for: self.header) { header in
```
kyleve: Consider just changing this to a `do { ... }` block?

kyleve: Going to leave this.

Listable/Sources/Internal/PresentationState.swift
```diff
@@ -415,51 +386,86 @@ final class PresentationState
...
+        }
+        
+        func applyToVisibleView()
+        {
+            guard let view = visibleContainer?.content, let state = self.state else {
+                return
+            }
             
-            self.cellRegistrationInfo = (class: SupplementaryItemView<Element>.self, reuseIdentifier: self.model.reuseIdentifier.stringValue)
+            state.applyTo(view: view, reason: .wasUpdated)
```
kyleve: self.state

Listable/Sources/Internal/SupplementaryContainerView.swift
```diff
@@ -0,0 +1,141 @@
+//
+//  SupplementaryContainerView.swift
+//  Listable
+//
+//  Created by Kyle Van Essen on 9/30/19.
+//
+
+import UIKit
+
+final class SupplementaryContainerView : UICollectionReusableView
```
kyleve: Comment what this is for

kyleve: Done

PR #97, opened 2020-03-04
Rename wasUpdated to isEquivalent

kyleve:
Rename wasUpdated to isEquivalent, and invert bool handling... More similar to implementing Equatable, which makes it less surprising.

### DIFF
--- a/BlueprintLists/Sources/HeaderFooter.swift
+++ b/BlueprintLists/Sources/HeaderFooter.swift
@@ -87,9 +87,9 @@ public struct BlueprintHeaderFooterElementAppearance : HeaderFooterElementAppear
     
     public func apply(to view: ContentView) {}
     
-    public func wasUpdated(comparedTo other: BlueprintHeaderFooterElementAppearance) -> Bool
+    public func isEquivalent(to other: BlueprintHeaderFooterElementAppearance) -> Bool
     {
-        return false
+        return true
     }
 }
 
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -5,8 +5,8 @@ PODS:
   - BlueprintLists/Tests (0.1.1):
     - BlueprintUI
     - Listable
-  - BlueprintUI (0.3.0)
-  - BlueprintUICommonControls (0.3.1):
+  - BlueprintUI (0.6.0)
+  - BlueprintUICommonControls (0.6.0):
     - BlueprintUI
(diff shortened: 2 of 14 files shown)


BlueprintLists/Sources/HeaderFooter.swift
```diff
@@ -87,9 +87,9 @@ public struct BlueprintHeaderFooterElementAppearance : HeaderFooterElementAppear
     
     public func apply(to view: ContentView) {}
     
-    public func wasUpdated(comparedTo other: BlueprintHeaderFooterElementAppearance) -> Bool
+    public func isIdenticalTo(comparedTo other: BlueprintHeaderFooterElementAppearance) -> Bool
```
kylebshr: Not to be too nit-picky, but the double "to" reads weirdly to me (`isIdenticalToComparedTo`) - maybe it should be `isIdentical(comparedTo other:)` or `isIdentical(to other:)`? Or even `isIdenticalTo(_ other:)`. Naming is hard.

kylebshr: FWIW `isEqual(_:)` is how Swift does that but I don't like it

Podfile.lock
```diff
@@ -28,7 +28,7 @@ DEPENDENCIES:
   - Snapshot/Tests (from `Internal Pods/Snapshot/Snapshot.podspec`)
 
 SPEC REPOS:
-  https://github.com/CocoaPods/Specs.git:
+  trunk:
```
kylebshr: Is this change intentional?

BlueprintLists/Sources/HeaderFooter.swift
kyleve: Ah yeah, good point. Workflow uses `public func isEquivalent(to otherWorker: Self) -> Bool`, which is also not bad. I'll think about it for a bit.

kyleve: Gonna do `sEquivalent(to:)`.

kyleve: Pushed!

Podfile.lock
kyleve: This was from updating BlueprintUI – I believe because I havent updated it since I updated Cocoapods, this changes it to pull from the spec CDN, not the repo.

Issue #96, opened 2020-03-03
Support pinning list to bottom as new rows are added


kyleve:
This allows implementing of the SPOS cart experiences (newest line items added at bottom, always visible), as well as other things like messaging apps.

kylebshr:
I think there are two (or more?) features to support here and want to make sure I'm working on ones we need first. 

1. Insetting the top content inset when the content height is less than the view height, so items are always at the bottom of the screen (Messaging apps)
2. Scrolling to the bottom of the list when a new item is added (which I think is what SPOS needs?)

I assume SPOS doesn't want the behavior of (1). Let me know if I'm wrong here!

kyleve:
Correct, yeah, we'll only want #2 right now! Good catch.

kyleve:
(#1 is actually already there for underflow behaviour, actually, which I didnt realize until I typed that last comment)

kylebshr:
I whipped up a demo view controller that just uses the existing `scrollTo` API to scroll to the last item after it's added. It works decently well. I noticed that the offset doesn't seem to be used at all, since we're forwarding to `scrollToItem`, which doesn't support an offset. 

https://github.com/kyleve/Listable/pull/101

kyleve:
Oh huh, im surprised that works as well as it does! Nice.

(You're right on the offset – I never moved over to using an API that lets us provide offsets.)

kylebshr:
I'll look into supporting the offset, though it probably makes this 10x more complicated 😆might need that overview first

kyleve:
I think we should be able to do that via using `setContentOffset(_:animated:)` instead, and then using the `collectionViewLayout`'s `layoutAttributesForItem(at:)` to get the position – hopefully not super complex... I think

kyleve:
(Also, fwiw, thinking about this more, it would be nice to expose pin to bottom as a config option on the list – so its just a bool for callers to set, vs additional behaviour.)

kylebshr:
I totally agree, I started with a property when I was thinking about the API and then noticed this and wanted to see if it worked. Having to just set it would be great though.

kyleve:
One thing I was thinking about yesterday, to control if a row should be scrolled to, we could add a `ScrollBehaviour` to `Item`, which could be `.none, always`, so we could control when an inserted row should be scrolled to.

kyleve:
(That also has the benefit of supporting arbitrary scrolling, not just pinning)

PR #101, opened 2020-03-05
Add AutoScrollingBehavior with option to scroll to item on insert

kylebshr:
This PR adds an enum for configuring the list view with `AutoScrollingBehavior`. It has two cases: `none` and `scrollToItemOnInsert(_:, position:)`.

When the list is updated, if `scrollToItemOnInsert` is set, the list checks if the item in the enum case was added in this update. If it was, it scrolls to the item after reloading the content. You can see this in the attached GIF:

Kapture 2020-04-01 at 11 35 04
Fixes https://github.com/kyleve/Listable/issues/96

### DIFF
--- a/Listable/Sources/AutoScrollAction.swift
+++ b/Listable/Sources/AutoScrollAction.swift
@@ -1,5 +1,5 @@
 //
-//  AutoScrollingBehavior.swift
+//  AutoScrollAction.swift
 //  Listable
 //
 //  Created by Kyle Bashour on 3/30/20.
@@ -8,7 +8,7 @@
 import Foundation
 
 
-public enum AutoScrollingBehavior
+public enum AutoScrollAction
 {
     /// The list never automatically scrolls.
     case none
--- a/Listable/Sources/AutoScrollAction.swift
+++ b/Listable/Sources/AutoScrollAction.swift
@@ -8,8 +8,8 @@
 import Foundation
 
 
-public enum AutoScrollAction
-{
+/// Options for auto-scrolling to items when the list is updated.
+public enum AutoScrollAction {
     /// The list never automatically scrolls.
     case none
 
--- a/Listable/Sources/AutoScrollAction.swift
+++ b/Listable/Sources/AutoScrollAction.swift
@@ -13,6 +13,6 @@ public enum AutoScrollAction
     /// The list never automatically scrolls.
     case none
 
-    /// Scrolls to the specified item when the list is updated.
-    case scrollToItemOnInsert(_ item: AnyItem, position: ItemScrollPosition)
+    /// Scrolls to the specified item when the list is updated if the item was inserted in this update.
+    case scrollToItemOnInsert(_ item: AnyItem, position: ItemScrollPosition, animated: Bool)
 }
--- a/Listable/Sources/Behavior.swift
+++ b/Listable/Sources/Behavior.swift
@@ -11,9 +11,14 @@ import Foundation
 public struct Behavior : Equatable
 {
     public var dismissesKeyboardOnScroll : Bool
+    public var pinItemsToBottom : Bool
     
-    public init(dismissesKeyboardOnScroll : Bool = false)
+    public init(
+        dismissesKeyboardOnScroll : Bool = false,
+        pinItemsToBottom : Bool = false
+    )
     {
         self.dismissesKeyboardOnScroll = dismissesKeyboardOnScroll
+        self.pinItemsToBottom = pinItemsToBottom
     }
 }
--- a/Listable/Sources/Behavior.swift
+++ b/Listable/Sources/Behavior.swift
@@ -11,14 +11,9 @@ import Foundation
 public struct Behavior : Equatable
 {
     public var dismissesKeyboardOnScroll : Bool
-    public var pinItemsToBottom : Bool
     
-    public init(
-        dismissesKeyboardOnScroll : Bool = false,
-        pinItemsToBottom : Bool = false
-    )
+    public init(dismissesKeyboardOnScroll : Bool = false)
     {
         self.dismissesKeyboardOnScroll = dismissesKeyboardOnScroll
-        self.pinItemsToBottom = pinItemsToBottom
     }
 }
--- a/Listable/Sources/ListDescription.swift
+++ b/Listable/Sources/ListDescription.swift
@@ -14,7 +14,7 @@ public struct ListDescription
     
     public var appearance : Appearance
     public var behavior : Behavior
-    public var autoScrollingBehavior : AutoScrollingBehavior
+    public var autoScrollAction : AutoScrollAction
     public var scrollInsets : ScrollInsets
     
     public var content : Content
@@ -25,7 +25,7 @@ public struct ListDescription
         animatesChanges: Bool,
         appearance : Appearance,
         behavior : Behavior,
-        autoScrollingBehavior : AutoScrollingBehavior,
+        autoScrollAction : AutoScrollAction,
         scrollInsets : ScrollInsets,
         build : Build
     )
@@ -33,7 +33,7 @@ public struct ListDescription
         self.animatesChanges = animatesChanges
         self.appearance = appearance
         self.behavior = behavior
-        self.autoScrollingBehavior = autoScrollingBehavior
+        self.autoScrollAction = autoScrollAction
         self.scrollInsets = scrollInsets
         
         self.content = Content()
--- a/Listable/Sources/ListDescription.swift
+++ b/Listable/Sources/ListDescription.swift
@@ -14,6 +14,7 @@ public struct ListDescription
     
     public var appearance : Appearance
     public var behavior : Behavior
+    public var autoScrollingBehavior : AutoScrollingBehavior
     public var scrollInsets : ScrollInsets
     
     public var content : Content
@@ -24,6 +25,7 @@ public struct ListDescription
         animatesChanges: Bool,
         appearance : Appearance,
         behavior : Behavior,
+        autoScrollingBehavior : AutoScrollingBehavior,
         scrollInsets : ScrollInsets,
         build : Build
     )
@@ -31,6 +33,7 @@ public struct ListDescription
         self.animatesChanges = animatesChanges
         self.appearance = appearance
         self.behavior = behavior
+        self.autoScrollingBehavior = autoScrollingBehavior
         self.scrollInsets = scrollInsets
         
         self.content = Content()
--- a/Listable/Sources/ListView/ListView.swift
+++ b/Listable/Sources/ListView/ListView.swift
@@ -156,21 +156,8 @@ public final class ListView : UIView
         // Nothing right now.
     }
 
-    public var autoScrollingBehavior : AutoScrollingBehavior {
-        didSet {
-            guard oldValue != self.autoScrollingBehavior else {
-                return
-            }
-
-            self.applyAutoScrollingBehavior()
-        }
-    }
-
-    private func applyAutoScrollingBehavior()
-    {
-        // Nothing right now.
-    }
-
+    public var autoScrollingBehavior : AutoScrollingBehavior
+    
     public var scrollInsets : ScrollInsets {
         didSet {
             guard oldValue != self.scrollInsets else {
@@ -216,12 +203,20 @@ public final class ListView : UIView
         guard let toIndexPath = self.storage.allContent.indexPath(for: item) else {
             return false
         }
+
+        // Check if the item is visible using its frame, since `visibleIndexPaths` includes items outside of the actual content frame.
         
+        let isAlreadyVisible: Bool = {
+            guard let frame = self.layout.layoutAttributesForItem(at: toIndexPath)?.frame else {
+                return false
+            }
+
+            return self.collectionView.contentFrame.contains(frame)
+        }()
+
         // If the item is already visible and that's good enough, return.
-        
-        let isAlreadyVisible = self.collectionView.indexPathsForVisibleItems.contains(toIndexPath)
-        
-        if  isAlreadyVisible && position.ifAlreadyVisible == .doNothing {
+
+        if isAlreadyVisible && position.ifAlreadyVisible == .doNothing {
             return true
         }
         
@@ -592,7 +587,7 @@ public final class ListView : UIView
             switch self.autoScrollingBehavior {
             case .none:
                 visibleSlice = self.storage.allContent.sliceTo(indexPath: indexPath, plus: Content.Slice.defaultSize)
-            case .scrollToBottomForNewItems:
+            case .scrollToItemOnInsert:
                 visibleSlice = Content.Slice(containsAllItems: true, content: self.storage.allContent)
             }
         }
@@ -620,13 +615,10 @@ public final class ListView : UIView
             callerCompletion(finished)
         }
 
-        // TODO: this is horrrrible
-        let oldLastIndex = IndexPath(item: (diff.old.last?.items.count ?? 1) - 1, section: diff.old.count - 1)
-        let newLastIndex = IndexPath(item: (diff.new.last?.items.count ?? 1) - 1, section: diff.new.count - 1)
-        let isNewItemAppended = (newLastIndex.section > oldLastIndex.section || newLastIndex.item > oldLastIndex.item) && diff.changes.itemsChangeCount == 1
-
-        if isNewItemAppended && self.autoScrollingBehavior == .scrollToBottomForNewItems {
-            self.scrollToBottom()
+        if case let AutoScrollingBehavior.scrollToItemOnInsert(autoScrollItem, autoScrollPosition) = self.autoScrollingBehavior,
+            !diff.old.contains(item: autoScrollItem), diff.new.contains(item: autoScrollItem)
+        {
+            self.scrollTo(item: autoScrollItem, position: autoScrollPosition, animated: true)
         }
 
         // Update info for new contents.
@@ -853,5 +845,12 @@ fileprivate extension UIScrollView
             self.alwaysBounceVertical = false
             self.alwaysBounceHorizontal = bounce
         }
+    }
+}
+
+extension Array where Element == Section {
+    func contains(item: AnyItem) -> Bool {
+        return self.flatMap { $0.items }
+            .contains { $0.identifier == item.identifier }
     }
 }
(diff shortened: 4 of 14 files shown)


Listable/Sources/ListView/ListView.swift
```diff
@@ -229,7 +229,47 @@ public final class ListView : UIView
...
+            }
+
+            let contentOffsetY = contentHeight - contentFrameHeight -
+                self.collectionView.lst_adjustedContentInset.top
+            let contentOffset = CGPoint(x: self.collectionView.contentOffset.x, y: contentOffsetY)
+            self.collectionView.setContentOffset(contentOffset, animated: animated)
+        }
+
+        if lastLoadedIndexPath < toIndexPath {
+            self.updatePresentationState(for: .programaticScrollDownTo(toIndexPath)) { _ in
```
kyleve: Since this is duplicated across the scroll item and this method, can we centralize the "I need to update the slice" bit?

Listable/Sources/Behavior.swift
```diff
@@ -11,9 +11,14 @@ import Foundation
 public struct Behavior : Equatable
 {
     public var dismissesKeyboardOnScroll : Bool
+    public var pinItemsToBottom : Bool
```
kyleve: staysPinnedToLastItem? staysScrolledToLastItem?

kylebshr: The more I think about this the more I think we should only scroll to the bottom if an item is _inserted_ at the bottom, right? If you delete an item at the bottom, the standard `performUpdates` will handle this, and if you insert/delete/move anything else, you probably don't want this to happen. If you agree, then maybe the name should reflect that - `scrollsToBottomOnInsertion` or something

Listable/Sources/ListView/ListView.swift
```diff
@@ -577,6 +625,17 @@ public final class ListView : UIView
             self.updateVisibleItemsAndSections()
             callerCompletion(finished)
         }
+
+        // Perform any other behavior-related changes.
+
+        switch reason {
+        case .contentChanged:
```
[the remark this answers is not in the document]
kylebshr: Probably, I'll try that!

Listable/Sources/Behavior.swift
kylebshr: `pinBottomInserts` `pinInsertionsToBottom`

kyleve: `pinsToBottomForNewItems`?

kyleve:
@kylebshr is this worth reviewing again now?

Listable/Sources/Behavior.swift
kyleve: Bump

Listable/Sources/ListView/ListView.swift
```diff
@@ -217,23 +213,39 @@ public final class ListView : UIView
...
-            self.updatePresentationState(for: .programaticScrollDownTo(toIndexPath)) { _ in
-                scroll()
+    }
+
+    @discardableResult
+    public func scrollToBottom(animated : Bool = true) -> Bool {
+
+        // Make sure we have a valid last index path.
+
+        guard let toIndexPath = self.storage.allContent.lastIndexPath() else {
```
kyleve, line 230: We have this, and then we also have `lastIndexPath` on `PresentationState` – can the `PresentationState` one defer through to this?

Listable/Sources/ListView/ListView.swift
```diff
@@ -521,6 +533,10 @@ public final class ListView : UIView
         case .contentChanged:
             self.updateCollectionViewConfiguration()
             self.updatePresentationStateWith(firstVisibleIndexPath: indexPath, for: reason, completion: completion)
+
+            if self.behavior.pinItemsToBottom {
```
kyleve, line 537: I think we should only do this if we inserted an item, right? I think we need to think through this behaviour a bit more. Eg, we may also want to remove the `Behaviour` level change, and instead tag items as `scrollToWhenInserted` or similar.

Listable/Sources/ListView/ListView.swift
kylebshr: Probably! Will look into this

Listable/Sources/Behavior.swift
kylebshr: Sorry for the long lapse on this, will be wrapping up some changes and pushing today! `pinsToBottomForNewItems` seems good to me

Listable/Sources/ListView/ListView.swift
kylebshr: > I think we should only do this if we inserted an item, right?

Yep, agreed, though the two qualms I have with making it an item level change are:

- More work for the consumer when they just want pinToBottom, since they have to figure out which items to add the flag to (not actually that hard though)
- Scrolling to an item doesn't have the same behavior as pin to bottom, as it doesn't account for safe area and content insets

Listable/Sources/ListView/ListView.swift
kylebshr: I think the reason I needed this is because `PresentationState` may not have all the items in it, so the `lastIndexPath` isn't the true last one of the list - is that assumption correct?

kyleve: Ahh true

Listable/Sources/ListView/ListView.swift
```diff
@@ -200,38 +203,58 @@ public final class ListView : UIView
         guard let toIndexPath = self.storage.allContent.indexPath(for: item) else {
             return false
         }
+
+        // Check if the item is visible using its frame, since `visibleIndexPaths` includes items outside of the actual content frame.
```
kylebshr, line 207: This fixes a bug where scrolling to an item wouldn't work for `doNothing` if it was 1-2 items offscreen.

Listable/Sources/ListView/ListView.swift
```diff
@@ -215,23 +227,39 @@ public final class ListView : UIView
...
             )
         }
-        
-        if lastLoadedIndexPath < toIndexPath {
-            self.updatePresentationState(for: .programaticScrollDownTo(toIndexPath)) { _ in
-                scroll()
+    }
+
+    @discardableResult
+    public func scrollToBottom(animated : Bool = true) -> Bool {
```
kylebshr, line 235: Thinking we should keep this since it seems generally useful, should I move to another PR?

Listable/Sources/ListView/ListView.swift
```diff
@@ -578,12 +614,39 @@ public final class ListView : UIView
             self.updateVisibleItemsAndSections()
             callerCompletion(finished)
         }
-        
+
+        if case let AutoScrollingBehavior.scrollToItemOnInsert(autoScrollItem, autoScrollPosition) = self.autoScrollingBehavior,
```
kylebshr, line 618: I couldn't figure out a nice way to keep this in the switch above, since I need the switch to calculate the diff, and need the diff to do this. If anyone sees a more elegant solution let me know

Listable/Sources/ListDescription.swift
```diff
@@ -14,6 +14,7 @@ public struct ListDescription
     
     public var appearance : Appearance
     public var behavior : Behavior
+    public var autoScrollingBehavior : AutoScrollingBehavior
```
kyleve, line 17: Maybe `autoScrollAction`?

Listable/Sources/ListView/ListView.swift
kyleve: Ooh, good catch

Listable/Sources/ListView/ListView.swift
```diff
@@ -154,6 +155,8 @@ public final class ListView : UIView
     {
         // Nothing right now.
     }
+
+    public var autoScrollingBehavior : AutoScrollingBehavior
```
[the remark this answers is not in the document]
kyleve: (It's of course not bad to have it here, just thinking out loud)

Listable/Sources/ListView/ListView.swift
```diff
@@ -553,11 +578,22 @@ public final class ListView : UIView
...
+        let visibleSlice: Content.Slice
+
+        if self.bounds.isEmpty {
+            visibleSlice = Content.Slice()
+        } else {
+            switch self.autoScrollingBehavior {
+            case .none:
+                visibleSlice = self.storage.allContent.sliceTo(indexPath: indexPath, plus: Content.Slice.defaultSize)
+            case .scrollToItemOnInsert:
+                visibleSlice = Content.Slice(containsAllItems: true, content: self.storage.allContent)
```
kyleve, line 591: This should probably only slice to the index of the item if it's contained, right?

Listable/Sources/ListView/ListView.swift
```diff
@@ -578,12 +614,39 @@ public final class ListView : UIView
             self.updateVisibleItemsAndSections()
             callerCompletion(finished)
         }
-        
+
+        if case let AutoScrollingBehavior.scrollToItemOnInsert(autoScrollItem, autoScrollPosition) = self.autoScrollingBehavior,
+            !diff.old.contains(item: autoScrollItem), diff.new.contains(item: autoScrollItem)
```
kyleve, line 619: Wondering if it's easier / faster to instead check the inserted items in the diff – since this could be comparing a _lot_ of items.

Listable/Sources/ListView/ListView.swift
```diff
@@ -578,12 +614,39 @@ public final class ListView : UIView
...
+        {
+            self.scrollTo(item: autoScrollItem, position: autoScrollPosition, animated: true)
```
kyleve, line 621: Should animated always be true here?

Listable/Sources/ListView/ListView.swift
```diff
@@ -768,3 +847,10 @@ fileprivate extension UIScrollView
         }
     }
 }
+
+private extension Array where Element == Section {
+    func contains(item: AnyItem) -> Bool {
+        return self.flatMap { $0.items }
```
kyleve, line 853: Wondering about the performance of this for large sections / lists. What about a nested `contains(where:)` instead?

Listable/Sources/ListView/ListView.swift
```diff
@@ -200,38 +203,58 @@ public final class ListView : UIView
...
+            guard contentHeight > contentFrameHeight else {
+                return
             }
-        } else {
-            scroll()
+
+            let contentOffsetY = contentHeight - contentFrameHeight -
+                self.collectionView.lst_adjustedContentInset.top
+            let contentOffset = CGPoint(x: self.collectionView.contentOffset.x, y: contentOffsetY)
+            self.collectionView.setContentOffset(contentOffset, animated: animated)
```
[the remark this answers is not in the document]
kylebshr: `updatePresentationStateAndScroll` is probably a confusing method name.. it takes a function to do the scrolling, which is what is being implemented here. Will clean up.

Listable/Sources/ListView/ListView.swift
kylebshr: ah yeah, meant to change

Listable/Sources/ListView/ListView.swift
kylebshr: _Is_ there an inserted items I'm missing? I couldn't find one 🤔

Listable/Sources/ListView/ListView.swift
kylebshr: Maaybe not I was thinking it would be but no reason not to make it configurable

Listable/Sources/ListView/ListView.swift
kylebshr: Probably better!

Listable/Sources/ListView/ListView.swift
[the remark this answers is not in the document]
kylebshr: Alternative would be to pass through `setProperties` → `setContent` → `set` → `setContentFromSource` → `updatePresentationState` - property seems simpler to me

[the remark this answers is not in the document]
kyleve: Sounds good!

Listable/Sources/ListView/ListView.swift
kyleve: Kind of, but it's a level below where you'd expect because there is first a diff of sections which then contain changes to individual items. I was thinking something like this: https://github.com/kyleve/Listable/compare/kve/insertedIDs?expand=1

Listable/Sources/ListView/ListView.swift
kyleve: Thinking eg, you set up a list with no animation because you're about to push it on screen, we wouldn't want this to have animation

Listable/Sources/ListView/ListView.swift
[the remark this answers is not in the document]
kyleve: Ahhh got it, right

Listable/Sources/ListView/ListView.swift
kylebshr: Will actually remove this in favor of checking in the diff like you suggested above

Listable/Sources/AutoScrollAction.swift
```diff
@@ -0,0 +1,18 @@
...
+
+import Foundation
+
+
+public enum AutoScrollAction
+{
+    /// The list never automatically scrolls.
+    case none
+
+    /// Scrolls to the specified item when the list is updated.
```
kyleve, line 16: Nit: Comment should mention that it only happens on insert

Listable/Sources/ListView/ListView.swift
```diff
@@ -553,11 +578,23 @@ public final class ListView : UIView
...
+
+        let visibleSlice: Content.Slice
+
+        if self.bounds.isEmpty {
+            visibleSlice = Content.Slice()
+        } else {
+            switch self.autoScrollAction {
+            case .none:
+                visibleSlice = self.storage.allContent.sliceTo(indexPath: indexPath, plus: Content.Slice.defaultSize)
+            case .scrollToItemOnInsert(let autoScrollItem, _):
```
kyleve, line 590: I don't think there's a great solution here, but it would be nice if we only use the `autoScrollItem`'s index path when performing the scroll. Wonder if `Content.Slice.UpdateReason` could be leveraged for this?

Listable/Sources/ListView/ListView.swift
```diff
@@ -553,11 +578,23 @@ public final class ListView : UIView
...
+                let indexPath = self.storage.allContent.indexPath(for: autoScrollItem.identifier) ?? indexPath
+                visibleSlice = self.storage.allContent.sliceTo(indexPath: indexPath, plus: Content.Slice.defaultSize)
```
kyleve, line 592: This needs to be either this `indexPath`, OR the `firstVisibleIndexPath`, whichever is larger.

Listable/Sources/ListView/ListView.swift
kylebshr: Updated! Blatantly stole your code, but added some tests - I think it works as we need it to.

Listable/Sources/ListView/ListView.swift
```diff
@@ -200,38 +203,58 @@ public final class ListView : UIView
...
+
+        // Make sure we have a valid last index path.
+
+        guard let toIndexPath = self.storage.allContent.lastIndexPath() else {
+            return false
+        }
+
+        // Perform scrolling.
+
+        return self.updatePresentationStateForScroll(toIndexPath: toIndexPath)  {
```
kyleve, line 245: Was thinking, maybe `extendPresentationState(to indexPath : IndexPath) { ... }`?

kyleve: Or `preparePresentationStateForScroll(to:)`?

Listable/Sources/ListView/ListView.swift
kylebshr: Hmm, we could set it back to `none` if it was `scrollToItemOnInsert`, which means it would just use the `firstVisibleIndexPath` next time unless `autoScrollAction` was set again. Does that seem like a good fix?

Listable/Tests/Internal/Diff/SectionedDiffTests.swift
```diff
@@ -13,3 +13,116 @@ class SectionedDiffTests: XCTestCase
...
+
+class SectionedDiff_SectionChangesTests: XCTestCase
+{
+
+    struct Section {
+        let identifier: Int
+        let items: [Int]
+    }
+
+    func test_added_items_in_added_section() {
```
kyleve, line 25: Test structure nit: use `func testcase("description") { ... }` within a single test method to group related tests?

Listable/Sources/ListView/ListView.swift
kyleve: I think so – this is one reason I was thinking it doesnt make sense to store `autoScrollAction` as a property, since it really is more of an action to perform during the current update cycle vs. to persist on the view itself. Hmmm

kylebshr: Yeah good point.. I'm gonna play with piping it through the functions

Listable/Sources/ListView/ListView.swift
```diff
@@ -578,12 +621,39 @@ public final class ListView : UIView
...
+            self.scrollTo(item: autoScrollItem, position: autoScrollPosition, animated: animated)
+        }
+
         // Update info for new contents.
         
         self.updateCollectionViewSelections(animated: reason.animated)
     }
-        
+
+    private func preparePresentationStateForScroll(toIndexPath: IndexPath, scroll: @escaping () -> Void) -> Bool {
```
kyleve, line 636: `preparePresentationStateForScroll(to indexPath: IndexPath`? Usually don't encode type in Swift var names, IIRC.

Listable/Sources/ListView/ListView.swift
```diff
@@ -154,6 +155,8 @@ public final class ListView : UIView
     {
         // Nothing right now.
     }
+
+    public var autoScrollAction : AutoScrollAction
```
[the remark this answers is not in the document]
kylebshr: Making this private - I did take a stab at removing the property, but it was pretty convoluted to pass it around right now. Would love to discuss with you separately from this PR though, since I could see us need to pass other "actions" like this that shouldn't actually be stored

Issue #100, opened 2020-03-05
Remove safe area inset when calculating keyboard inset


kylebshr:
It seems like the keyboard content inset adjustment still has the safe area inset, which results in a gap at the bottom of the scroll view when scrolled all the way down while the keyboard is up. Most obvious when comparing to a device without a safe area inset.

| iPhone 11 |  iPhone 8 |
|-|-|
| ![Simulator Screen Shot - iPhone 11 - 2020-03-05 at 10 37 40](https://user-images.githubusercontent.com/3526783/76013876-8459b680-5ecd-11ea-9c0f-e3543d9948ed.png) | ![Simulator Screen Shot - iPhone 8 - 2020-03-05 at 10 38 50](https://user-images.githubusercontent.com/3526783/76013886-87ed3d80-5ecd-11ea-9053-f0bef4e9be6b.png) |

Issue #99, opened 2020-03-05
Set scroll indicator insets when content insets are adjusted


kylebshr:
When content insets are set (for example, when the keyboard is showing) the scroll indicator insets should be set to match

PR #85, opened 2019-12-14
Fix usage of contentInset on iOS 10 and earlier.

kyleve:

PR #103, opened 2020-03-06
Fix content inset issues with scroll indicators and keyboard insets

kylebshr:
- Fixes https://github.com/kyleve/Listable/issues/100
- Fixes https://github.com/kyleve/Listable/issues/99

This PR fixes an issue with the scroll indicator insets not being set to match the content insets, which is almost always the correct behavior. In the first GIF you can see the scroll indicator disappears behind the keyboard, while the in the second gif it matches the content size. 

This PR also fixes an issue when calculating the inset to add to adjust for a visible keyboard. I chose to subtract the bottom content inset from the calculation, as the collection view keeps the safe area inset in addition to insets you set. Another option might be to disable safe area inset adjustments entirely, but then we would lose the automatic adjustment when in most cases we want it. In the first GIF you can see the gap between the collection view content and the keyboard. In the second, there's no gap. Both have the correct safe area when the keyboard is not visible.

| Before | After |
|-|-|
| Kapture 2020-03-06 at 10 07 32 | Kapture 2020-03-06 at 10 05 11 |

### DIFF
--- a/Listable/Sources/ListView/ListView.swift
+++ b/Listable/Sources/ListView/ListView.swift
@@ -719,11 +719,17 @@ extension ListView : KeyboardObserverDelegate
             return
         }
         
-        var inset : CGFloat
+        let inset : CGFloat
         
         switch frame {
-        case .notVisible: inset = 0.0
-        case .visible(let frame): inset = (self.bounds.size.height - frame.origin.y)
+        case .notVisible:
+            inset = 0.0
+        case .visible(let frame):
+            if #available(iOS 11, *) {
+                inset = (self.bounds.size.height - frame.origin.y) - self.safeAreaInsets.bottom
+            } else {
+                inset = (self.bounds.size.height - frame.origin.y)
+            }
         }
         
         self.collectionView.contentInset.bottom = inset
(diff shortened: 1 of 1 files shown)


kyleve:
Any idea if this also resolves https://github.com/kyleve/Listable/pull/85 ?

kylebshr:
It's not super clear to me what the issue was for #85, what's the incorrect behavior on <= iOS 10?

kyleve:
> It's not super clear to me what the issue was for #85, what's the incorrect behavior on <= iOS 10?

On iOS 10, the scroll view sets the contentInset for tab bar or nav bar insets – because iOS 7+ underflowed those controls, but safe area wasnt a thing until iOS 11. Reading this, I dont think this makes it any better or worse, so, yay

kyleve:
Also, I never even knew `scrollIndicatorInsets` was a thing. TIL!

kylebshr:
Yah, I dunno why it’s not just automatically set TBH. Nice detail to fix though, always bugs me when apps with custom headers and stuff don’t (Spotify comes to mind)

PR #107, opened 2020-03-31
Raise BlueprintLists iOS target to 10.0

watt:


kyleve:
LG, but also should we do Listable while we're at it?

watt:
> should we do Listable while we're at it?

Left that for another pass, there are some `available` checks to look at.

PR #110, opened 2020-04-03
[WIP] Swipe to action v1

mattfaluotico:
*Started on top of Kyles Skeleton commit

Adds support for simple swipe to actions 

why simple? 
- no custom views
- only one action

What's going on? 

Created a SwipeController. The SwipeController manages the gesture of the cell itself and the state of the swipe. It helps calculate new states and fire the correct action. When actions are sent into a cell instance, a SwipeController and swipe view are created. 

*SwipeView* - a simple view that mimics the action view from UITableViewCell. 

The movement is handled with a UIPanGestureRecognizer added to the cell itself. 

-- video

ezgif com-video-to-gif

### DIFF
--- a/Demo/Sources/DemosRootViewController.swift
+++ b/Demo/Sources/DemosRootViewController.swift
@@ -129,12 +129,15 @@ public final class DemosRootViewController : UIViewController
                     with: DemoItem(text: "Swipe To Action"),
                     selection: .isSelectable(isSelected: false),
                     swipeActions: SwipeActions(SwipeAction(
-                    title: "Delete",
-                    backgroundColor: .purple,
-                    image: nil, onTap: { _ in
-                        print("Delete")
-                        return true
-                    }), performsFirstOnFullSwipe: false),
+                        title: "Delete",
+                        backgroundColor: .purple,
+                        image: nil,
+                        onTap: { _ in
+                            print("Deleted")
+                            return true
+                        }
+                    ),performsFirstOnFullSwipe: true),
+                    swipeActionsAppearance: DefaultItemElementSwipeActionsAppearance(),
                     onSelect : { _ in
                         self.push(DemoTableViewController())
                 })
--- a/Demo/Sources/DemosRootViewController.swift
+++ b/Demo/Sources/DemosRootViewController.swift
@@ -134,7 +134,7 @@ public final class DemosRootViewController : UIViewController
                     image: nil, onTap: { _ in
                         print("Delete")
                         return true
-                    }), performsFirstOnFullSwipe: true),
+                    }), performsFirstOnFullSwipe: false),
                     onSelect : { _ in
                         self.push(DemoTableViewController())
                 })
--- a/Demo/Sources/DemosRootViewController.swift
+++ b/Demo/Sources/DemosRootViewController.swift
@@ -118,6 +118,20 @@ public final class DemosRootViewController : UIViewController
                         self.push(FlowLayoutViewController())
                 })
             }
+            
+            list += Section(identifier: "table-view") { section in
+                
+                section.header = HeaderFooter(
+                    with: DemoHeader(title: "Table Views")
+                )
+                
+                section += Item(
+                    with: DemoItem(text: "Swipe To Action"),
+                    selection: .isSelectable(isSelected: false),
+                    onSelect : { _ in
+                        self.push(DemoTableViewController())
+                })
+            }
         }
     }
 }
--- a/Demo/Sources/DemosRootViewController.swift
+++ b/Demo/Sources/DemosRootViewController.swift
@@ -128,6 +128,13 @@ public final class DemosRootViewController : UIViewController
                 section += Item(
                     with: DemoItem(text: "Swipe To Action"),
                     selection: .isSelectable(isSelected: false),
+                    swipeActions: SwipeActions(SwipeAction(
+                    title: "Delete",
+                    backgroundColor: .purple,
+                    image: nil, onTap: { _ in
+                        print("Delete")
+                        return true
+                    }), performsFirstOnFullSwipe: true),
                     onSelect : { _ in
                         self.push(DemoTableViewController())
                 })
--- a/Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
+++ b/Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
@@ -1,0 +1,141 @@
+//
+//  DefaultItemElementSwipeActionsAppearance.swift
+//  AppHost-BlueprintLists-Unit-Tests
+//
+//  Created by Matthew Faluotico on 4/14/20.
+//
+
+import Foundation
+
+public final class DefaultItemElementSwipeActionsAppearance: ItemElementSwipeActionsAppearance {
+
+    public init() { }
+
+}
+
+// MARK: - ItemElementSwipeActionsAppearance
+public extension DefaultItemElementSwipeActionsAppearance {
+
+    static func createView(frame: CGRect) -> SwipeView {
+        return .init()
+    }
+
+    func apply(swipeActions: SwipeActions, to view: SwipeView) {
+        // Currently only supporting first action
+        if let action = swipeActions.actions.first {
+            view.action = action
+        }
+    }
+
+    func apply(swipeControllerState: SwipeControllerState, to view: DefaultItemElementSwipeActionsAppearance.SwipeView) {
+        view.state = swipeControllerState
+    }
+
+    func preferredSize(for view: SwipeView) -> CGSize {
+        return view.preferredSize()
+    }
+
+}
+
+// MARK: - SwipeView
+extension DefaultItemElementSwipeActionsAppearance {
+
+    public final class SwipeView: UIView {
+
+        private static var padding: UIEdgeInsets
+        {
+            return UIEdgeInsets(top: 16, left: 16, bottom: 16, right: 16)
+        }
+
+        var state: SwipeControllerState
+        {
+            didSet {
+                UIView.animate(withDuration: 0.2) {
+                    switch self.state {
+                    case .pending, .swiping(passedSwipeThroughThreshold: false), .locked:
+                        self.stackView.alignment = .center
+                    case .swiping(passedSwipeThroughThreshold: true), .finished:
+                        self.stackView.alignment = .leading
+                    }
+                }
+            }
+        }
+
+        var action: SwipeAction? {
+            didSet {
+                self.titleView?.text = action?.title
+                self.backgroundColor = action?.backgroundColor ?? .white
+
+                if let image = action?.image {
+                    if let imageView = self.imageView {
+                        imageView.image = image
+                    } else {
+                        let imageView = UIImageView(image: image)
+                        self.stackView.addArrangedSubview(imageView)
+                        self.imageView = imageView
+                    }
+                }
+            }
+        }
+
+        private var imageView: UIImageView?
+        private var titleView: UILabel?
+        private var gestureRecognizer: UITapGestureRecognizer
+        private var stackView = UIStackView()
+
+        init()
+        {
+            self.state = .pending
+            self.gestureRecognizer = UITapGestureRecognizer()
+            super.init(frame: .zero)
+
+            self.gestureRecognizer.addTarget(self, action: #selector(onTap))
+            self.addGestureRecognizer(gestureRecognizer)
+
+            let titleView = UILabel()
+            titleView.textColor = .white
+            titleView.lineBreakMode = .byClipping
+            self.stackView.addArrangedSubview(titleView)
+            self.titleView = titleView
+
+            stackView.spacing = 2
+            stackView.alignment = .center
+            stackView.axis = .vertical
+            stackView.distribution = .fill
+            stackView.isLayoutMarginsRelativeArrangement = true
+            stackView.layoutMargins = SwipeView.padding
+
+            self.addSubview(self.stackView)
+        }
+
+        required init?(coder: NSCoder)
+        {
+            fatalError("init(coder:) has not been implemented")
+        }
+
+        public override func layoutSubviews()
+        {
+            super.layoutSubviews()
+            self.stackView.frame = self.bounds
+        }
+
+        func preferredSize() -> CGSize
+        {
+            guard let titleView = self.titleView else {
+                return self.sizeThatFits(UIScreen.main.bounds.size)
+            }
+
+            var size = titleView.sizeThatFits(self.bounds.size)
+            size.width += SwipeView.padding.left + SwipeView.padding.right
+            return size
+        }
+
+        @objc func onTap()
+        {
+            if let action = self.action {
+                let _ = action.onTap(action)
+            }
+        }
+    }
+
+}
--- a/Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
+++ b/Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
@@ -11,27 +11,24 @@ public final class DefaultItemElementSwipeActionsAppearance: ItemElementSwipeAct
 
     public init() { }
 
-}
-
-// MARK: - ItemElementSwipeActionsAppearance
-public extension DefaultItemElementSwipeActionsAppearance {
-
-    static func createView(frame: CGRect) -> SwipeView {
+    public static func createView(frame: CGRect) -> SwipeView {
         return .init()
     }
 
-    func apply(swipeActions: SwipeActions, to view: SwipeView) {
-        // Currently only supporting first action
+    public func apply(swipeActions: SwipeActions, to view: SwipeView) {
+        guard swipeActions.actions.count <= 1 else {
+            fatalError("More than one action is not currently supported")
+        }
         if let action = swipeActions.actions.first {
             view.action = action
         }
     }
 
-    func apply(swipeControllerState: SwipeControllerState, to view: DefaultItemElementSwipeActionsAppearance.SwipeView) {
-        view.state = swipeControllerState
+    public func apply(swipeControllerState: SwipeControllerState, to view: DefaultItemElementSwipeActionsAppearance.SwipeView) {
+        view.setState(swipeControllerState, animated: true)
     }
 
-    func preferredSize(for view: SwipeView) -> CGSize {
+    public func preferredSize(for view: SwipeView) -> CGSize {
         return view.preferredSize()
     }
 
@@ -44,59 +41,44 @@ extension DefaultItemElementSwipeActionsAppearance {
 
         private static var padding: UIEdgeInsets
         {
-            return UIEdgeInsets(top: 16, left: 16, bottom: 16, right: 16)
+            return UIEdgeInsets(top: 0, left: 16, bottom: 0, right: 16)
         }
 
-        var state: SwipeControllerState
-        {
-            didSet {
-                UIView.animate(withDuration: 0.2) {
-                    switch self.state {
-                    case .pending, .swiping(passedSwipeThroughThreshold: false), .locked:
-                        self.stackView.alignment = .center
-                    case .swiping(passedSwipeThroughThreshold: true), .finished:
-                        self.stackView.alignment = .leading
-                    }
-                }
-            }
-        }
+        private (set) var state: SwipeControllerState
 
         var action: SwipeAction? {
             didSet {
-                self.titleView?.text = action?.title
+                self.titleView.text = action?.title
                 self.backgroundColor = action?.backgroundColor ?? .white
 
                 if let image = action?.image {
-                    if let imageView = self.imageView {
-                        imageView.image = image
-                    } else {
-                        let imageView = UIImageView(image: image)
-                        self.stackView.addArrangedSubview(imageView)
-                        self.imageView = imageView
+                    self.imageView.image = image
+
+                    if !stackView.arrangedSubviews.contains(imageView) {
+                        stackView.addArrangedSubview(imageView)
                     }
                 }
             }
         }
 
-        private var imageView: UIImageView?
-        private var titleView: UILabel?
-        private var gestureRecognizer: UITapGestureRecognizer
-        private var stackView = UIStackView()
+        lazy private var imageView = UIImageView()
+        private let titleView: UILabel
+        private let gestureRecognizer: UITapGestureRecognizer
+        private let stackView = UIStackView()
 
         init()
         {
-            self.state = .pending
+            self.state = .closed
             self.gestureRecognizer = UITapGestureRecognizer()
+            self.titleView = UILabel()
             super.init(frame: .zero)
 
             self.gestureRecognizer.addTarget(self, action: #selector(onTap))
             self.addGestureRecognizer(gestureRecognizer)
 
-            let titleView = UILabel()
             titleView.textColor = .white
             titleView.lineBreakMode = .byClipping
             self.stackView.addArrangedSubview(titleView)
-            self.titleView = titleView
 
             stackView.spacing = 2
             stackView.alignment = .center
@@ -121,10 +103,6 @@ extension DefaultItemElementSwipeActionsAppearance {
 
         func preferredSize() -> CGSize
         {
-            guard let titleView = self.titleView else {
-                return self.sizeThatFits(UIScreen.main.bounds.size)
-            }
-
             var size = titleView.sizeThatFits(self.bounds.size)
             size.width += SwipeView.padding.left + SwipeView.padding.right
             return size
@@ -134,6 +112,21 @@ extension DefaultItemElementSwipeActionsAppearance {
         {
             if let action = self.action {
                 let _ = action.onTap(action)
+            }
+        }
+
+        public func setState(_ state: SwipeControllerState, animated: Bool)
+        {
+            self.state = state
+            if animated {
+                UIView.animate(withDuration: 0.2) {
+                    switch self.state {
+                    case .closed, .swiping(passedSwipeThroughThreshold: false), .open:
+                        self.stackView.alignment = .center
+                    case .swiping(passedSwipeThroughThreshold: true), .finished:
+                        self.stackView.alignment = .leading
+                    }
+                }
             }
         }
     }
--- a/Listable/Sources/Internal/ItemElementCell.ContentView.swift
+++ b/Listable/Sources/Internal/ItemElementCell.ContentView.swift
@@ -1,0 +1,45 @@
+//
+//  ItemCellView.swift
+//  Listable
+//
+//  Created by Kyle Van Essen on 3/23/20.
+//
+
+import UIKit
+
+
+extension ItemElementCell
+{
+    final class ContentView : UIView
+    {
+        private(set) var contentScrollView : UIScrollView
+        
+        private(set) var contentView : Element.Appearance.ContentView
+        
+        private(set) var swipeView : Element.SwipeActionsAppearance.ContentView
+        
+        override init(frame : CGRect)
+        {
+            let bounds = CGRect(origin: .zero, size: frame.size)
+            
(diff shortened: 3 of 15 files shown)


kyleve:
FYI, attached GIF does not seem to be loading.

Listable/Sources/Internal/ItemElementCell.ContentView.swift
```diff
@@ -0,0 +1,135 @@
...
+
+            let width = self.bounds.width
+            var frame = self.contentView.frame
+            frame.origin.x = newOrigin_x
+            self.contentView.frame = frame
+
+            let originX = frame.maxX
+            let swipeWidth = width - originX
+
+            self.swipeView?.frame = CGRect(
```
kyleve, line 74: Rather than doing this this way, I think UIViewPropertyAnimator would be a better choice for these cases – you pause it while the gesture recognizer is active, and scrub the completion amount manually, and then re-activate the animator when it's not active to complete the animation.

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    // MARK: - Panning
+
+    @objc func onPan(_ pan: UIPanGestureRecognizer)
+    {
+        switch pan.state {
+        case .began:
+            guard let collectionView = SwipeController.findCollectionView(child: containerView), self.collectionView == nil else { return }
```
kyleve, line 82: I don't love walking up the view hierarchy to do this – generally superview walking is a smell – is there a reason to put this gesture on the collection view vs. on each cell / content view?

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,285 @@
...
+import Foundation
+
+protocol SwipeControllerDelegate: class {
+
+    func swipeController(panDidMove controller: SwipeController)
+    func swipeController(panDidEnd controller: SwipeController)
+
+}
+
+class SwipeController: NSObject {
```
kyleve, line 17: `final class`; also, I don't believe this needs to inherit from NSObject because you're using `@objc` below

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,285 @@
...
+                        self.stackView.alignment = .leading
+                    }
+                }
+            }
+        }
+
+        private var action: SwipeAction
+
+        private var imageView: UIImageView?
+        private var titleView: UILabel!
```
kyleve, line 220: How come this is an implicitly unwrapped optional?

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        return nil
+
+    }
+}
+
+extension SwipeController {
+
+    class SwipeView: UIView {
```
kyleve, line 196: I don't love that we have visual behaviour baked in here – the reason for `ItemElementSwipeActionsAppearance` is to support arbitrary views. Should make this be an implementation detail of a standard `ItemElementSwipeActionsAppearance`.

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        private var action: SwipeAction
+
+        private var imageView: UIImageView?
+        private var titleView: UILabel!
+        private var gestureRecognizer: UITapGestureRecognizer!
+        private var stackView = UIStackView()
+
+        init(action: SwipeAction)
```
kyleve, line 224: Thought: You could also implement this with a BlueprintView rather than plain UIKit if it would be easier.

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,285 @@
...
+        func preferredSize() -> CGSize
+        {
+            var size = self.titleView.sizeThatFits(self.bounds.size)
+            size.width += SwipeView.padding.left + SwipeView.padding.right
+            return size
+        }
+
+        @objc func onTap()
+        {
+            let _ = self.action.onTap(action)
```
kyleve, line 281: If the return value isn't used here, it should likely be removed from the API. Note that the original reason for this boolean value was to indicate if the action should be performed, similar to how UIContextualAction works https://developer.apple.com/documentation/uikit/uicontextualaction/handler

Listable/Sources/ItemElement.swift
```diff
@@ -146,6 +151,52 @@ public extension ItemElementAppearance where Self:Equatable
...
+/**
+ TODO
+ */
+public protocol ItemElementSwipeActionsAppearance
+{
+    //
+    // MARK: Creating & Providing Views
+    //
+    
+    /// TODO
```
kyleve, line 163: Need some comments

Listable/Sources/Internal/SwipeController.swift
kyleve: (This would allow the gesture recognizer, etc, since you could just use a Column in a Tappable to handle the actions)

kyleve: (In this case though, you'd want it to be in an Appearance in SPOS, not in Listable, since there's no BP dependency here)

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: So this gesture (i should add a few comments here) pushes the cell back into place when the user starts scrolling normally again.

if you swipe a mail item to the locking point (where it shows the action items) and then start scrolling the collection view again, it pushes back into it's normal place. 

Technically, it's a bit more optional. You could just keep the cell in it's swipe state when the collection view scrolls.

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: I agree. 

This was more of a v1 / default-appearance approach. 

Would `ItemElementSwipeActionsAppearance` own the default view?

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: - [ ] final class
- [ ] remove NSObject

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: I was under the assumption all actions would have some form of title. 

it should not be though. Will remove.

mattfaluotico: - [ ] Better Optional unwrapping

Listable/Sources/ItemElement.swift
mattfaluotico: - [ ] comments

kyleve:
I was thinking about the "only one action" bit – I think you'll need multiple, at least for GA, because comp & void requires two actions. Right? cc @connorcimowsky

mattfaluotico:
> I was thinking about the "only one action" bit – I think you'll need multiple, at least for GA, because comp & void requires two actions. Right? cc @connorcimowsky

Chatted with Connor offline, he said there only needs to be one

mattfaluotico:
If the plan for a more complete implementation is to support custom UIView / Blueprint Action appearances, the underlying swipe view will need to communicate a few things to the ContentView that holds the pan gesture. 

1. How many actions there are
2. Preferred size for each view 

The ContentView's locked position (see attache screenshot in reminders) is calculated based on these two properties. 

If the complete implementation returns one UIView, containing all actions views, the apply function would need to bind the action to the respective action view. 

IMG_F7F52FE7502E-1

Listable/Sources/Internal/SwipeController.swift
kyleve: Yeah – it would – or rather, an object which conforms to it – so that this default / basic v1 is not baked into the core implementation (but instead, just one possible implementation).

Listable/Sources/Internal/ItemElementCell.ContentView.swift
mattfaluotico: @kylebshr thoughts given the issues on iOS 10?

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: The Bool is to inform the collection/table view that the action was performed successfully, right? 

Internally, we don't do anything different when this returns false vs true, but in the future we may want to?

Listable/Sources/Internal/PresentationState.swift
```diff
@@ -640,17 +642,25 @@ final class PresentationState
...
             self.model.element.apply(
-                to: cell.content,
+                to: cell.content.contentView,
                 for: reason,
                 with: applyInfo
             )
+            
+            // Apply Swipe To Action Appearance
+            
+            if let actions = self.model.swipeActions, let appearance = self.model.swipeActionsAppearance {
```
kylebshr, line 659: I wonder if we could model these as a single config item since they're both required to work? Or assert if `actions` is set without an appearance.

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,210 @@
...
+
+    func swipeControllerPanDidMove()
+    func swipeControllerPanDidEnd()
+
+}
+
+public enum SwipeControllerState {
+
+    // Docked position without the action view showing
+    case pending
```
kylebshr, line 20: I'd propose `closed` instead of `pending` and `open` instead of `locked`.

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,210 @@
...
+    // Pan was released and Actions are visible occupying their required space from preferredSize()
+    case locked
+    // Pan has swiped all the way to the end, only when isSwipeThroughEnabled is true
+    case finished
+
+}
+
+final class SwipeController<Appearance: ItemElementSwipeActionsAppearance> {
+
+    var state: SwipeControllerState = .pending {
```
kylebshr, line 32: `private(set)`?

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,210 @@
...
+//  Listable
+//
+//  Created by Matthew Faluotico on 4/3/20.
+//
+
+import Foundation
+
+protocol SwipeControllerDelegate: class {
+
+    func swipeControllerPanDidMove()
```
kylebshr, line 12: I think you should pass the gesture recognizer through these methods, or the entire swipe controller

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,210 @@
...
+
+    var state: SwipeControllerState = .pending {
+        didSet {
+            if let swipeView = self.swipeView {
+                appearance.apply(swipeControllerState: state, to: swipeView)
+            }
+        }
+    }
+
+    private (set) var actions: SwipeActions
```
kylebshr, line 40: Just wondering if all these properties need to be exposed, or if some of them can be fully `private`

Listable/Sources/Internal/SwipeController.swift
kylebshr: Ah, you're not actually adding a pan gesture to the collection view, you're using the built in one to detect a scroll.. clever. I also wish we could avoid walking up the hierarchy, but without moving the implementation for swiping into `ListView` I don't think it's really possible. You could add a GR to the window, but that might conflict with scrolling and other gestures, so this seems like the best solution to me

Podfile.lock
```diff
@@ -28,7 +28,7 @@ DEPENDENCIES:
   - Snapshot/Tests (from `Internal Pods/Snapshot/Snapshot.podspec`)
 
 SPEC REPOS:
-  trunk:
+  https://cdn.cocoapods.org/:
```
kylebshr, line 31: Should this get merged? Changes back to `trunk` when I run pod install

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
```diff
@@ -0,0 +1,141 @@
...
+
+        private static var padding: UIEdgeInsets
+        {
+            return UIEdgeInsets(top: 16, left: 16, bottom: 16, right: 16)
+        }
+
+        var state: SwipeControllerState
+        {
+            didSet {
+                UIView.animate(withDuration: 0.2) {
```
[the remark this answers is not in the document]
kyleve: +1 – setters should always be unanimated, and pass through to the `setFoo(animated::)` method

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: Moved this implementation into a Default Apperance

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: Default View uses Gesture, but appearance can be built with anything otherwise.

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// MARK: - ItemElementSwipeActionsAppearance
+public extension DefaultItemElementSwipeActionsAppearance {
+
+    static func createView(frame: CGRect) -> SwipeView {
+        return .init()
+    }
+
+    func apply(swipeActions: SwipeActions, to view: SwipeView) {
```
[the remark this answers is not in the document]
mattfaluotico: yeah. That was what was thinking. There's a decent amount of customization that would need to go down

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: Passing the swipe controller gets messy because it's a generic. Any advice on that one?

Podfile.lock
mattfaluotico: yeah, I'll fix that.

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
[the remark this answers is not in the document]
mattfaluotico: - [x] `setState(animated:)`

Podfile.lock
mattfaluotico: - [ ] fix that

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: - [x] check property visibility

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: - [x] yes private (set)

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: yeah I like that more too. 
- [ ] closed
- [ ] open

Listable/Sources/Internal/ItemElementCell.ContentView.swift
```diff
@@ -0,0 +1,138 @@
...
+            guard swipeController != nil else { return }
+
+            swipeController = nil
+            swipeView?.removeFromSuperview()
+            swipeView = nil
+        }
+
+        public func registerSwipeActionsIfNeeded(actions: SwipeActions, appearance: Element.SwipeActionsAppearance)
+        {
+            guard self.swipeController == nil else { return } // Already Registered
```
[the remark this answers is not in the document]
mattfaluotico: I think that's a good call

- [ ] allow apply actions to call

Listable/Sources/Internal/ItemElementCell.ContentView.swift
```diff
@@ -0,0 +1,138 @@
...
+//
+//  Created by Kyle Van Essen on 3/23/20.
+//
+
+import UIKit
+
+
+extension ItemElementCell
+{
+    final class ContentView : UIView, SwipeControllerDelegate
```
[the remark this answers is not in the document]
mattfaluotico: cc @kyleve, as this is his commit. Thoughts?

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
```diff
@@ -0,0 +1,141 @@
...
+        public override func layoutSubviews()
+        {
+            super.layoutSubviews()
+            self.stackView.frame = self.bounds
+        }
+
+        func preferredSize() -> CGSize
+        {
+            guard let titleView = self.titleView else {
+                return self.sizeThatFits(UIScreen.main.bounds.size)
```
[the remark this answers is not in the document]
mattfaluotico: nah, I think zero is a better move. 
- [ ] use `.zero` instead of UIScreen.

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
```diff
@@ -0,0 +1,141 @@
...
+                        self.imageView = imageView
+                    }
+                }
+            }
+        }
+
+        private var imageView: UIImageView?
+        private var titleView: UILabel?
+        private var gestureRecognizer: UITapGestureRecognizer
+        private var stackView = UIStackView()
```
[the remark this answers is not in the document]
mattfaluotico: - [ ] let

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
```diff
@@ -0,0 +1,141 @@
...
+                        self.stackView.addArrangedSubview(imageView)
+                        self.imageView = imageView
+                    }
+                }
+            }
+        }
+
+        private var imageView: UIImageView?
+        private var titleView: UILabel?
+        private var gestureRecognizer: UITapGestureRecognizer
```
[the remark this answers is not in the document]
mattfaluotico: - [ ] let

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
```diff
@@ -0,0 +1,141 @@
...
+                        let imageView = UIImageView(image: image)
+                        self.stackView.addArrangedSubview(imageView)
+                        self.imageView = imageView
+                    }
+                }
+            }
+        }
+
+        private var imageView: UIImageView?
+        private var titleView: UILabel?
```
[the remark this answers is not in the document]
mattfaluotico: - [ ] let

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
```diff
@@ -0,0 +1,141 @@
...
+                    } else {
+                        let imageView = UIImageView(image: image)
+                        self.stackView.addArrangedSubview(imageView)
+                        self.imageView = imageView
+                    }
+                }
+            }
+        }
+
+        private var imageView: UIImageView?
```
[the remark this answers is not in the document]
mattfaluotico: - [ ] investigate lazy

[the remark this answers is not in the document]
mattfaluotico: we would need to be clever about adding it to the stackview.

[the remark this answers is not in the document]
mattfaluotico: Something like this:

```
if let image = action?.image {
                    self.imageView.image = image

                    if !stackView.arrangedSubviews.contains(imageView) {
                        stackView.addArrangedSubview(imageView)
                    }
                }
```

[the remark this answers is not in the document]
mattfaluotico: yeah , I think i prefer the lazy implementation.

Listable/Sources/Internal/DefaultItemElementSwipeActionsAppearance.swift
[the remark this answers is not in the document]
mattfaluotico: was able to eliminate this by removing the optionality on titleView

Listable/Sources/Internal/ItemElementCell.ContentView.swift
```diff
@@ -0,0 +1,138 @@
...
+
+            let swipeController = SwipeController<Element.SwipeActionsAppearance>(
+                appearance: appearance,
+                swipeView: swipeView,
+                actions: actions,
+                contentView: self.contentView,
+                containerView: self
+            )
+
+            swipeController.configure()
```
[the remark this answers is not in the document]
mattfaluotico: I was avoiding adding the gesture recognizer to the container view when you initialize the swipe controller.

Listable/Sources/Internal/PresentationState.swift
mattfaluotico: I think throwing an assert makes sense. Harder to diagnose errors without one. What do you think @kyleve ?

Listable/Sources/Internal/SwipeController.swift
mattfaluotico: removed

Listable/Sources/Internal/PresentationState.swift
kyleve: I kept them separate in here since they're separate on the `Item` – and putting them into one type on `Item` felt like the callsite ended up being too cumbersome, IMO:

```
struct SwipeSettings<AppearanceType:ItemElementSwipeActionsAppearance> {
   ...
}

swipe: SwipeSettings(
    actions: .init(...)
    appearance: Appearance
)

```

It's not terrible, but I just didnt like it. Anyone else feel differently?

Fine with either an assert, or just failing out early and not showing anything

mattfaluotico: I like keeping them separate, since it mimics `Item`.  I can add an assert.

mattfaluotico: - [ ] Add assert

Listable/Sources/Internal/ItemElementCell.ContentView.swift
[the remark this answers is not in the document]
kylebshr: I'd say go for it, he asked me to drive this forward so we don't have too many cooks :)

Demo/Sources/DemosRootViewController.swift
```diff
@@ -125,6 +125,30 @@ public final class DemosRootViewController : UIViewController
...
+            list += Section(identifier: "table-view") { section in
+                
+                section.header = HeaderFooter(
+                    with: DemoHeader(title: "Table Views")
+                )
+                
+                section += Item(
+                    with: DemoItem(text: "Swipe To Action"),
+                    selection: .isSelectable(isSelected: false),
+                    swipeActions: SwipeActions(SwipeAction(
```
kyleve, line 138: Hah, including this on this row is smart

Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
```diff
@@ -0,0 +1,141 @@
...
+//
+//  Created by Kyle Van Essen on 3/23/20.
+//
+
+import UIKit
+
+
+extension ItemElementCell
+{
+    final class ContentContainerView : UIView, SwipeControllerDelegate
```
kyleve, line 13: Nit: The name of the class is backwards from the name of then file, and `//  ItemCellView.swift` in the comments at the top were not updated.

Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
```diff
@@ -0,0 +1,141 @@
...
+    {
+        private(set) var contentView : Element.Appearance.ContentView
```
kyleve, line 15: Can this be `let`?

Listable/Sources/Internal/PresentationState.swift
```diff
@@ -454,6 +454,8 @@ final class PresentationState
         {
             let view = view as! Element.Appearance.ContentView
             
+            // TODO: Merge this with the other place we apply to views.
```
kyleve, line 457: Leftover TODO, remove?

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -0,0 +1,210 @@
...
+import Foundation
+
+protocol SwipeControllerDelegate: class {
+
+    func swipeControllerPanDidMove()
+    func swipeControllerPanDidEnd()
+
+}
+
+public enum SwipeControllerState {
```
kyleve, line 17: This enum is public, but is in the `Internal` directory – if it needs to remain public, we should move out of this file.

Listable/Sources/ItemElement.swift
```diff
@@ -38,6 +38,11 @@ public protocol ItemElement
      */
     associatedtype Appearance:ItemElementAppearance
     
+    /**
+
```
kyleve, line 42: Missing doc comment

kyleve:
Overall looks good – left some notes to clean up in a follow up.

kyleve:
Fwiw, I'm now noticing this auto layout issue in the console when scrolling the main demo app list:

```
2020-05-02 14:24:20.015508-0700 Demo[91896:4120979] [LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x600002b05270 'UISV-canvas-connection' UILayoutGuide:0x600003134620'UIViewLayoutMarginsGuide'.leading <= UILabel:0x7faeaad10b30'Delete'.leading   (active)>",
    "<NSLayoutConstraint:0x600002b052c0 'UISV-canvas-connection' UILayoutGuide:0x600003134620'UIViewLayoutMarginsGuide'.centerX == UILabel:0x7faeaad10b30'Delete'.centerX   (active)>",
    "<NSLayoutConstraint:0x600002b04e60 'UIView-Encapsulated-Layout-Width' UIStackView:0x7faeaad08740.width == 0   (active)>",
    "<NSLayoutConstraint:0x600002b04fa0 'UIView-leftMargin-guide-constraint' H:|-(16)-[UILayoutGuide:0x600003134620'UIViewLayoutMarginsGuide'](LTR)   (active, names: '|':UIStackView:0x7faeaad08740 )>",
    "<NSLayoutConstraint:0x600002b05040 'UIView-rightMargin-guide-constraint' H:[UILayoutGuide:0x600003134620'UIViewLayoutMarginsGuide']-(16)-|(LTR)   (active, names: '|':UIStackView:0x7faeaad08740 )>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x600002b05270 'UISV-canvas-connection' UILayoutGuide:0x600003134620'UIViewLayoutMarginsGuide'.leading <= UILabel:0x7faeaad10b30'Delete'.leading   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKitCore/UIView.h> may also be helpful.
```

The `Delete` in there leads me to believe it's related to this.

kylebshr:
Hmm might be because of the stack view, those use AL internally. My refactor completely changes how the buttons are laid out (to support multiple buttons) so it should fix this. I’ll double check, but I don’t think we need to bother looking into it

kyleve:
Cool, makes sense! Sorry for the weekend ping, was working on another side project that means I need to do some list layout refactoring, and ran into it so wanted to post.

kylebshr:
No worries! Wasn’t really even a ping, just happened to be browsing my email 😄

PR #123, opened 2020-05-03
Update types to pass sizes through instead of types

kyleve:


Listable/Sources/Internal/Layout/ListLayout.swift
```diff
@@ -227,7 +227,12 @@ public extension ListLayoutContent
         
         let isPopulated : Bool
         
-        var size : CGSize = .zero
+        var size : CGSize = .zero {
+            didSet {
+                precondition(self.size.height != .greatestFiniteMagnitude)
```
kyleve, line 232: Remove

Listable/Sources/Internal/PresentationState.swift
```diff
@@ -469,46 +469,50 @@ final class PresentationState
...
+                let size : CGSize = measurementCache.use(
                     with: self.model.reuseIdentifier,
                     create: {
                         return Element.Appearance.createReusableHeaderFooterView(frame: .zero)
                 }, { view in
                     self.model.appearance.apply(to: view)
                     self.model.element.apply(to: view, reason: .willDisplay)
                     
-                    return self.model.sizing.measure(with: view, width: width, layoutDirection: layoutDirection, defaultHeight: defaultHeight)
+                    return self.model.sizing.measure(with: view, in: sizeConstraint, layoutDirection: layoutDirection, defaultSize: defaultSize)
```
kyleve, line 510: This has been a longstanding bug, but the sizing should be taken into account in the cache as well.

kyleve: Nvm, handled by `setNew`

Listable/Sources/ListView/ListView.swift
```diff
@@ -215,26 +215,25 @@ public final class ListView : UIView
...
         
-        let isAlreadyVisible: Bool = {
-            guard let frame = self.layout.layoutAttributesForItem(at: toIndexPath)?.frame else {
-                return false
-            }
+        return self.preparePresentationStateForScroll(to: toIndexPath) {
+            
+            // Check if the item is visible using its frame, since `visibleIndexPaths` includes items outside of the actual content frame.
+            
+            let isAlreadyVisible: Bool = {
```
kyleve, line 223: @kylebshr: FYI, updating this method a bit. Since the passed in `indexPath` may be from an unloaded row, the `isAlreadyVisible` check needs to be nested within the `self.preparePresentationStateForScroll` call.

PR #129, opened 2020-05-04
Support a shouldPerform block for auto scrolling

kyleve:
Adds support for conditionally accepting or rejecting the `AutoScrollAction`, based on the currently rendered state of the table view.

1) Added a `CHANGELOG.md` similar to Blueprint.

1) Updated `AutoScrollingViewController` to utilize new APIs.

1) Added a `ScrollViewEdgesPlaygroundViewController` for testing scroll view behaviour.

1) Changed `scrollToItemOnInsert(..)` to `scrollToItem(onInsertOf: ...)`, which now takes in a single param of `OnInsertedItem`, which contains all properties. This makes using the enum much less wiedy.

1) Similarly, I changed the `scrollToItem(onInsertOf: ...)` to take in an `AnyIdentifier`, instead of an `AnyItem`, since this is what really matters to the API. There's a static method to make construction easier:

```
    public static func scrollTo(
        _ destination : ScrollDestination? = nil,
        onInsertOf insertedIdentifier: AnyIdentifier,
        position: ItemScrollPosition,
        animated: Bool = false,
        shouldPerform : @escaping (ListScrollPositionInfo) -> Bool = { _ in true }
    ) -> AutoScrollAction
    {
        .scrollToItem(onInsertOf: .init(
            destination: destination ?? .item(insertedIdentifier),
            insertedIdentifier: insertedIdentifier,
            position: position,
            animated: animated,
            shouldPerform: shouldPerform
        ))
    }
```

1) Introduce `ListScrollPositionInfo`, which is the "meat" of this PR. It provides a snapshot of the current state of the list, including the first visible item, the last visible item, all the visible items, and which edges of the list are currently visible (top, left, bottom, right).

1) Move all item and supplementary visibility control onto `VisibleContent`.

1) Fix bugs around visibility of items due to using `indexPathsForVisible...` methods on table view, switch to using the layout itself. Add tests.


Listable/Sources/AutoScrollAction.swift
```diff
@@ -14,5 +14,65 @@ public enum AutoScrollAction {
...
+            position: position,
+            animated: animated,
+            shouldPerform: shouldPerform
+        ))
+    }
+}
+
+public extension AutoScrollAction
+{
+    struct ScrollToItemOnInsert
```
kyleve, line 38: Nested these into a type since the long-winded enums get unwieldy, IMO.

Listable/Sources/AutoScrollAction.swift
```diff
@@ -14,5 +14,65 @@ public enum AutoScrollAction {
...
+        case top
+        case middle
+        case bottom
+    }
+    
+    public init(scrollView : UIScrollView, visibleItems : Set<AnyIdentifier>)
+    {
+        self.visibleItems = visibleItems
+        
+
```
kyleve, line 75: Need to implement the `Position` bit, haven't yet since it's actually hard and needs tests

Demo/Sources/CollectionView/AutoScrollingViewController.swift
```diff
@@ -61,7 +61,9 @@ final class AutoScrollingViewController : UIViewController
             list += Section(identifier: "items", items: items)
 
             if let last = items.last {
-                list.autoScrollAction = .scrollToItemOnInsert(last, position: .init(position: .bottom), animated: true)
+                list.autoScrollAction = .scrollToItemOnInsert(last.identifier, position: .init(position: .bottom), animated: true) { state in
+                    return true
```
kyleve, line 65: Eg in here you could check based on the passed in state to return true/false

Listable/Sources/AutoScrollAction.swift
```diff
@@ -14,5 +14,65 @@ public enum AutoScrollAction {
...
+        }
+    }
 }
+
+public struct ScrollState : Equatable {
+    
+    public var position : Position
+    public var visibleItems : Set<AnyIdentifier>
+    
+    public enum Position : Equatable {
```
kyleve, line 65: Actually, this doesnt make a ton of sense since we could see the entire view if the list is shorter than the view...

kyleve: Maybe 

```
public struct ScrollState : Equatable {
    
    public var isTopEdgeVisible : Bool
    public var isBottomEdgeVisible : Bool
    
    public var visibleItems : Set<AnyIdentifier>
}
```

Instead

DJBen: I agree that this makes more sense.

I am thinking of an `OptionSet`.
```swift
public struct VisibleEdges: OptionSet {
    public static let top = VisibleEdges(rawValue: 1)
    public static let bottom = VisibleEdges(rawValue: 1 >> 1)
}```
but I think it is an overkill. If two booleans could do.

Comment (kyleve) [reply Listable/Sources/AutoScrollAction.swift:65]: I like this – since technically this should work for left and right edges as well; since lists can be horiztonal. Will update!

Comment (kyleve) [reply Listable/Sources/AutoScrollAction.swift:65]: Actually, `UIRectEdge` would actually work very nicely for this

Comment (kyleve) [reply Listable/Sources/AutoScrollAction.swift]: Pushed an update for this

File: Listable/Sources/ListScrollPositionInfo.swift [hunk 5]
```diff
@@ -0,0 +1,170 @@
...
+        ///                                            +---------------+      +---------------+
+        /// ```
+        /// Safe Area Insets
+        /// -----------------
+        /// You can control whether `safeAreaInsets` should be taken into account via the `includingSafeAreaEdges` parameter.
+        ///
+        /// Generally, you want to include the `safeAreaInsets` for the top, left, and right, but may want to exclude the bottom
+        /// if you consider the bottom edge visible if it's visible below the home indicator on a home button-less iPhone or iPad.
+        ///
+        public func visibleContentEdges(includingSafeAreaEdges safeAreaEdges : UIRectEdge = .all) -> UIRectEdge
```
kyleve, line 86: - [ ] Test me

Listable/Sources/ListScrollPositionInfo.swift
```diff
@@ -0,0 +1,170 @@
...
+                safeAreaInsets: safeArea
+            )
+        }
+    }
+}
+
+
+extension UIRectEdge : CustomDebugStringConvertible
+{
+    static func visibleScrollViewContentEdges(
```
kyleve, line 120: - [ ] Test me

Listable/Sources/ListView/ListView.swift
```diff
@@ -644,10 +644,29 @@ public final class ListView : UIView
...
 
-        if case let AutoScrollAction.scrollToItemOnInsert(autoScrollItem, autoScrollPosition, animated) = self.autoScrollAction,
-            diff.changes.addedItemIdentifiers.contains(autoScrollItem.identifier)
+        if case let AutoScrollAction.scrollToItem(info) = self.autoScrollAction
         {
-            self.scrollTo(item: autoScrollItem, position: autoScrollPosition, animated: animated)
+            let wasInserted = diff.changes.addedItemIdentifiers.contains(info.identifier)
+            
+            if wasInserted {
+                let visibleItems = Set(self.visibleItems.map { item in
```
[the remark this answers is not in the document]
kyleve: Related: https://github.com/kyleve/Listable/pull/101#discussion_r401828718

Listable/Sources/ListView/ListView.swift
```diff
@@ -286,6 +308,13 @@ public final class ListView : UIView
             )
         }
     }
+    
+    @discardableResult
+    public func scrollToTop(animated : Bool = false) -> Bool {
```
kyleve, line 313: Test me

DJBen:
Woohoo thank you!

PR #130, opened 2020-05-06
Add support for multiple swipe actions and match UITableView more closely

kylebshr:
Swipe actions in `UITableView` offer a lot of magical behavior. This PR doesn't replicate the magic, but offers a more explicit API to accomplish everything you need to match UITableView's behavior. 

For context on the magical behavior, `UIContextualAction` has a completion handler that the consumer should call with a bool indicating whether or not the action was "completed". When this handler is called, the actions are closed regardless of the bool passed in. The exception is when you call `UITableView.deleteRows(at:with:)` while swipe actions are open - the table view automagically expands the actions, and animates a line matching the color of the action performed. This, again, happens regardless of the bool passed to the handler. 

I decided to give `SwipeAction` in Listable the same handler, but actually use the Bool to indicate whether the actions should expand (i.e., because the item was deleted) or just collapse back to closed. This is much more explicit, and still allows you to match the behavior of `UITableView` when you want to.

The following GIFs show the same feature created with Listable swipe actions and UITableView swipe actions:

| | Listable | UITableView |
|-|-|-|
| Action Interactions | Scrolling-Listable | Scrolling-Tableview |
| Destructive Action | Deleting-Listable | Deleting-Tableview |
| Non-destructive Action | Saving-Listable | Saving-Tableview |


Listable/Sources/Internal/UIViewProperyAnimtor+System.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+//  Created by Kyle Bashour on 4/17/20.
+//
+
+import UIKit
+
+extension UIViewPropertyAnimator {
+    convenience init(system animations: @escaping () -> Void) {
+        let params = UISpringTimingParameters()
+        self.init(duration: 1, timingParameters: params)
```
kyleve, line 13: Should 1 be hardcoded here? Probably better to pass it in

Listable/Sources/Item.swift
```diff
@@ -93,8 +92,7 @@ public struct Item<Element:ItemElement> : AnyItem
         sizing : Sizing = .default,
         layout : ItemLayout = ItemLayout(),
         selection : ItemSelection = .notSelectable,
-        swipeActions : SwipeActions? = nil,
-        swipeActionsAppearance : Element.SwipeActionsAppearance? = nil,
```
kyleve, line 97: I don't love that we're not passing in this any more – one reason for passing in an Appearance so you had a hook point to inject your own customizations, eg, market styling. Maybe we could pass in a closure initializing the `ItemElementSwipeActionsView`, with a default value that calls the normal `init(didPerformAction: @escaping SwipeAction.CompletionHandler)`?

Listable/Sources/ItemElement.swift
```diff
@@ -36,12 +36,12 @@ public protocol ItemElement
      If you have multiple elements with similar or identical appearances,
      they should likely share an ItemElementAppearance to reduce code duplication.
      */
-    associatedtype Appearance:ItemElementAppearance
+    associatedtype Appearance: ItemElementAppearance
     
     /**
```
kyleve, line 42: Not new, but, add a comment here?

Listable/Sources/ItemElement.swift
```diff
@@ -150,85 +150,33 @@ public extension ItemElementAppearance where Self:Equatable
     }
 }
 
+public protocol ItemElementSwipeActionsView: UIView {
```
kyleve, line 153: Needs documentation

Listable/Sources/SwipeActionsConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+//  SwipeActions.swift
+//  Listable
+//
+//  Created by Kyle Van Essen on 8/10/19.
+//
+
+import Foundation
+
+
+public struct SwipeActionsConfiguration {
```
kyleve, line 11: Needs documentation

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -590,12 +558,13 @@
...
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
+				DEVELOPMENT_TEAM = ZEAD23GG77;
 				INFOPLIST_FILE = Demo/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.kve.listable.Demo;
+				PRODUCT_BUNDLE_IDENTIFIER = com.kbe.listable.Demo;
```
kyleve, line 567: 👀

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -608,12 +577,13 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
+				DEVELOPMENT_TEAM = ZEAD23GG77;
```
kyleve, line 580: I think this should be reverted, since I assume it's a personal team

kyleve:
Also, probably worth a squash into one commit before merging master in

Listable/Sources/Internal/SwipeController.swift
```diff
@@ -1,210 +1,212 @@
+////
```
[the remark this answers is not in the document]
kylebshr: Oops good catch

Listable/Sources/Internal/UIViewProperyAnimtor+System.swift
kylebshr: This initializer ignores the duration completely, it's very dumb that you have to pass one in. I'll add a comment

kylebshr: (The duration is determined by the spring physics)

Listable/Sources/Item.swift
kylebshr: I guess I didn't really follow how the appearance object offered more customization - was the idea that you could subclass the `DefaultItemElementSwipeActionsAppearance` (or write your own) and customize the appearance in `createView` or `apply`?

Listable/Sources/Internal/DefaultSwipeView.swift
```diff
@@ -0,0 +1,156 @@
...
+private class DefaultSwipeActionButton: UIButton {
+
+    private let inset: CGFloat = 16
+    private var action: SwipeAction?
+    private var didPerformAction: ((Bool) -> Void)?
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+
+        titleLabel?.font = .systemFont(ofSize: 15, weight: .medium)
```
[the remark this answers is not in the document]
kylebshr: Yeah, there's a lot of places for customization and I wasn't sure what should be exposed where. I'll think on this a bit, it feels a little weird to have the action deciding things like font when it should be uniform across actions

Listable/Sources/ItemElement.swift
kylebshr: Will do - also been meaning to ask, maybe it makes sense to have `DefaultSwipeActionsView` actually be the default, and remove `EmptyItemElementSwipeActionsView`? The view isn't created unless actions are applied, so the empty one doesn't offer anything. Then swipe actions just work out of the box.

Demo/Demo.xcodeproj/project.pbxproj
kylebshr: oops

Demo/Demo.xcodeproj/project.pbxproj
kylebshr: yeppers

kyleve:
Also to test: How does this work with VoiceOver? We should test that as well.

-[ ] VoiceOver testing.

Note: Good doc from Alex here about some details: https://docs.google.com/document/d/1FKLvPKpUwiWu4mmKEEF50pbg1xBwlhdxvsCVVdyFQng/edit

Listable/Sources/Item.swift
kyleve: Agree – more calling out that since we're just "passing in" a type (via the associated type), there's no opportunity to customize the view with your own injected data. Maybe not important if we make the SwipeActions have a custom font, etc.

Listable/Sources/ItemElement.swift
kyleve: +1

kylebshr:
Looking into VoiceOver and there's some weird behavior in general. Some of the layouts don't seem to like navigating with VO very much. Cells also aren't accessibility items, so the content gets selected, which is a little weird. I pushed up a commit that makes the cell an a11y item, and adds a11y actions for the swipe actions to the cell. But that prevents the content from being selected by VO, which means the text isn't read out loud. Maybe ItemElement should have an accessibilityTitle property?

kyleve:
Ah interesting... Yeah, I guess we need an accessibility pass in general on Listable. Maybe dont worry about it for now, then?

Listable/Sources/Internal/DefaultSwipeView.swift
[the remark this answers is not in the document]
kyleve: Maybe it could be a property on the `Appearance` of the list?

Listable/Sources/Item.swift
kyleve: Resolving this for now, we can revisit

Listable/Sources/Internal/DefaultSwipeView.swift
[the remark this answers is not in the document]
kylebshr: Can I follow up in a separate PR? I'm realizing that you almost never want to customize the actual SwipeActionsView, but instead the swipe action (singular) view, and I think an appearance type for that makes sense

kylebshr:
@kyleve I think I addressed all the initial comments. The a11y solution I added is better than nothing IMO, thoughts on leaving it in?

Listable/Sources/Internal/DefaultSwipeView.swift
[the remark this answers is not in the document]
kyleve: Sounds good!

Demo/Sources/CollectionView/DemoTableViewController.swift
```diff
@@ -1,48 +0,0 @@
...
-//  Demo
-//
-//  Created by Kyle Van Essen on 3/24/20.
-//  Copyright © 2020 Kyle Van Essen. All rights reserved.
-//
-
-import UIKit
-
-
-final class DemoTableViewController : UITableViewController
```
kyleve, line 12: Not a big deal, but wouldn't mind keeping this to easily test other UITV stuff.

Listable/Sources/Internal/HorizontalPanGestureRecognizer.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  HorizontalPanGestureRecognizer.swift
+//  Listable
+//
+//  Created by Kyle Bashour on 4/21/20.
+//
+
+import UIKit.UIGestureRecognizerSubclass
+
+class HorizontalPanGestureRecognizer: UIPanGestureRecognizer {
```
kyleve, line 10: Nit: If this is only used within the swipe actions, I'd put it in that same file or nest it in that type? Not a big deal though.

Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
```diff
@@ -8,134 +8,259 @@
...
             super.init(frame: frame)
 
+            isAccessibilityElement = true
+
             self.addSubview(self.contentView)
+
+            NotificationCenter.default.addObserver(
+                self,
+                selector: #selector(didReceiveCloseNotification),
+                name: .closeSwipeActions, object: nil
```
kyleve, line 40: Nit: `object: nil` on a new line

Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
```diff
@@ -8,134 +8,259 @@
...
+            contentView.frame.origin.x = xOriginOffset
+            configuration.swipeView.frame = bounds.divided(atDistance: -xOriginOffset, from: .maxXEdge).slice
         }
 
         // MARK: - Swipe Registration
 
-        public func deregisterSwipeIfNeeded()
-        {
-            guard swipeController != nil else { return }
+        public func deregisterSwipeIfNeeded() {
```
kyleve, line 95: Does this need to be public? (Same with others in here)

Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
```diff
@@ -8,134 +8,259 @@
...
+                    self.setNeedsLayout()
+                    self.layoutIfNeeded()
+                }.startAnimation()
+            } else {
+                setNeedsLayout()
+            }
         }
 
-        // MARK: - Swipe Controller Delegate
+        @objc private func didReceiveCloseNotification(notification: Notification) {
```
kyleve, line 217: Nit: Do we want to check if `Notification.object` is one of our superviews? Guess it's also fine to just close all of them globally, though.

kylebshr:
Sounds good, will address these in my follow up for appearance!

PR #135, opened 2020-05-12
Selection & Highlight States + Background Views For Cells

kyleve:
**Changes**

1) I removed `ItemElementAppearance` and merged it down into `ItemElement`. While not critical to this PR, it made the addition of the `BackgroundView` and `SelectedBackgroundView` much easier. In reality, we haven't seen people use these `ItemElement` and `ItemElementAppearance` separately, and it made the code shape funny in a bunch of places. This let me delete more code than I added to add selection support.

1) Introduced support for background views and selected background views in both `ItemElement`, and `BlueprintItemElement`. 

1) I renamed `selection` and `ItemSelection` to `selectionStyle` and `ItemSelectionStyle`, respectively. I added a `tappable` case for temporary selection behaviour. **This was the main thing I was trying to accomplish in this PR.**

1) I removed the `with` from `Item(with: element)`, because it was already gone from the `BlueprintLists` extension, which was the primary use case, and makes the code read better.

1) Rename `createReusableItemView` to `createReusableContentView`, which now makes more sense since we have other views.

1) I added more doc comments for things that were missing.

1) Bumped Blueprint to 0.11.0, updated some demo project stuff to use the new element building APIs.

**Note: Since this is already a big PR, I will remove `HeaderFooterElementAppearance` in a follow up PR.**


Listable/Sources/Layout/ListLayout.swift
```diff
@@ -512,14 +512,14 @@ public extension ListLayoutContent
...
         func setContentsFrameWithContent() {
-            let allFrames : [CGRect] = [[
-                    self.header.defaultFrame,
-                    self.footer.defaultFrame
-                ],
-                self.items.map { $0.frame }
-                ].flatMap { $0 }
-            
-            //self.contentsFrame = .from(unioned: allFrames)
+//            let allFrames : [CGRect] = [[
```
kyleve, line 515: Commenting this out for now

Demo/Sources/CollectionViewAppearance.swift
```diff
@@ -92,18 +82,26 @@ struct DemoItem : BlueprintItemElement, Equatable
...
-                wrapping: Label(text: self.text)
-            )
+            borderStyle: .solid(color: .white(0.9), width: 2.0)
+        )
+    }
+    
+    func selectedBackgroundElement(with info: ApplyItemElementInfo) -> Element?
+    {
+        Box(
+            backgroundColor: info.state.isSelected ? .white(0.9) : .white(0.95),
```
kylebshr, line 101: Why are you checking if the state is selected and varying the styling? Won't the `backgroundElement` be displayed instead of this one if it's not selected? If that's not the case, I feel like this API is a little confusing.

kyleve: Somewhat weirdly, collection view's "selected background view" is also visible during highlighting, so this provides a lighter highlight during highlight vs. selection.

kyleve: (I was trying to come up with a better name for the API than "selected background", but didnt really come up with anything...)

kyleve: - [ ] Update documents to call this out

kyleve: Updated docs and readme.

PR #136, opened 2020-05-14
Fix Swipe Actions Bug

kyleve:
- Fix bug when close swipe actions notification was posted too often, and setter did not check state.

### DIFF
--- a/Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
+++ b/Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
@@ -202,6 +202,11 @@ extension ItemElementCell {
         }
 
         private func set(state: SwipeActionState, animated: Bool = false) {
+            
+            guard swipeState != state else {
+                return
+            }
+            
             swipeState = state
 
             if animated {
--- a/Listable/Sources/ListView/ListView.Delegate.swift
+++ b/Listable/Sources/ListView/ListView.Delegate.swift
@@ -360,6 +360,14 @@ extension ListView
         
         // MARK: UIScrollViewDelegate
         
+        func scrollViewWillBeginDragging(_ scrollView: UIScrollView)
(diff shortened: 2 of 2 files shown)


Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
```diff
@@ -202,6 +202,11 @@ extension ItemElementCell {
         }
 
         private func set(state: SwipeActionState, animated: Bool = false) {
```
kyleve, line 204: @kylebshr: One question I have is that I'm surprised this setter method and the property's didSet are different:

```
        private var swipeState: SwipeActionState = .closed {
            didSet {
                if oldValue != swipeState {
                    swipeConfiguration?.swipeView.apply(state: swipeState)
                }
            }
        }
```

Based on UIKit paradigm, I'd simply expect the `didSet` to call through to a `set(state: newValue, animated:false)`. If the behaviour is meant to be meaningfully different, should the name of this method be different?

kyleve:
Going to land this so I can continue with some other refactors that would conflict – still need to figure out the Q above

kyleve:
Ugh, wait, this may have broken the swipe animation, I think. Going to revert...

Listable/Sources/Internal/ItemElementCell.ContentViewContainer.swift
```diff
@@ -202,6 +202,11 @@ extension ItemElementCell {
...
+            
+            guard swipeState != state else {
```
kyleve, line 206: A ha, turns out this breaks the general case which causes the bug. Sometimes, this should always be called, and other times, it should only be called if the value changed.

kyleve:
Fix for this breaking ongoing swipe handling: https://github.com/kyleve/Listable/pull/140

PR #150, opened 2020-05-29
Rename "Element" to "Content"

kyleve:
The naming of `ItemElement`, `HeaderFooterElement` was a bit confusing, since it didn't really convey how it was different from `Item` or `HeaderFooter`, and also made talking about Blueprint integration confusing, because Blueprint refers to its content as `Element`. Thus, rename to `ItemContent` and `HeaderFooterContent`, plus update all relevant code, docs, comments, etc.


BlueprintLists/Sources/Section.swift
```diff
@@ -1,31 +0,0 @@
-//
```
kyleve, line 1: Moved into mainline.

BlueprintLists/Sources/HeaderFooter.swift
```diff
@@ -6,33 +6,28 @@
...
 
-public protocol BlueprintHeaderFooterElement : HeaderFooterElement where ContentView == BlueprintView
+public protocol BlueprintHeaderFooterContent : HeaderFooterContent where ContentView == BlueprintView
 {
     //
     // MARK: Creating Blueprint Element Representations
     //
     
-    var element : BlueprintUI.Element { get }
+    var elementRepresentation : Element { get }
```
kyleve, line 18: I changed this to line up with `ProxyElement`, so if you make a `ProxyElement` conform to `BlueprintHeaderFooterContent`, you don't also need an `element` property.

Listable/Sources/EmbeddedList.swift
```diff
@@ -1,39 +1,29 @@
...
-        switch self {
-        case .default: return .default
-        case .fixed(let w, let h): return .fixed(width: w, height: h)
-        }
-    }
-}
-
-
-public struct ListItemElement : ItemElement
+public struct EmbeddedList : ItemContent
```
kyleve, line 26: This name seemed a bit clearer, so went with this instead of `ListItemContent`.

Listable/Sources/EmbeddedList.swift
```diff
@@ -90,3 +80,19 @@ public struct ListItemElement : ItemElement
         ListView(frame: frame)
     }
 }
+
+public extension EmbeddedList
```
kyleve, line 84: Nested `ListItemSizing` into `EmbeddedList`.

Listable/Sources/Item/ItemContent.swift
```diff
@@ -209,8 +209,18 @@ public extension ItemElement where Self:Equatable
 }
 
 
+/// Provides a default implementation of `identifier` when self conforms to Swift's `Identifiable` protocol.
+@available(iOS 13.0, *)
+public extension ItemContent where Self:Identifiable
```
kyleve, line 214: Added this as well for easier `Identifiable` usage.

Listable/Sources/Section.swift
```diff
@@ -105,11 +105,16 @@ public struct Section
         lhs.add(rhs)
     }
     
-    public static func += <Element:ItemElement>(lhs : inout Section, rhs : Item<Element>)
+    public static func += <Content:ItemContent>(lhs : inout Section, rhs : Item<Content>)
     {
         lhs.add(rhs)
     }
     
+    public static func += <Content:ItemContent>(lhs : inout Section, rhs : Content)
```
kyleve, line 113: These came in from `BlueprintLists`.

Listable/Sources/Item/ItemContent.swift
kylebshr: I think `#if swift(>=5.1)` is more correct here - `Identifiable` should be available anywhere as long as it's Swift 5.1 right?

kyleve: Ahh yeah, maybe – let me try that; Xcode told me to add the iOS 13 check so I did

kyleve: Hmm, Xcode doesn't like it:

```
#if swift(>=5.1)
public extension ItemContent where Self:Identifiable
{
    var identifier : Identifier<Self> {
        .init(self.id)
    }
}
#endif
```

Still asks me to do:

```
/Users/k/Desktop/Development/Listable3/Listable/Sources/Item/ItemContent.swift:214:41: 'Identifiable' is only available in iOS 13 or newer
```

kylebshr: Dang, the docs didn't mention an iOS version so I thought it would work. Oh well!

kyleve: Might be because it was added in 5.1, which IIRC was the first ABI stable release, so it's now baked into iOS vs. embedded in the bin?

kyleve:
> Might not be worth it, but could make migration easier if you added type aliases for public renamed types and deprecated them. Something like:
> 
> ```swift
> @available(*, deprecated, renamed: "ItemContent")
> typealias ItemElement = ItemContent
> ```

I'd thought about doing this, but figured since we'll also be doing the bumping in SPOS, it wouldnt make a ton of diference – though I guess it could help people find the new APIs – thoughts?

kylebshr:
I know we're pre 1.0 but there's definitely consumers outside of SPOS already (i.e., that hacker news demo app, other side projects) so it'd be nice I think.

kyleve:
Good point! Will do

kyleve:
@kylebshr added some deprecations! https://github.com/kyleve/Listable/pull/150/files#diff-b9f5fdcf35b4a661e7012de6357de2e1

PR #156, opened 2020-06-03
Begin work on supporting KeyboardAdjustmentMode

kyleve:
When this lands in SPOS, go back and remove the need for manual overrides here: https://git.sqcorp.co/projects/IOS/repos/register/pull-requests/44979/


kyleve:
Don't forget to copy over the improvements from Blueprint's version of the keyboard observer here as well

PR #172, opened 2020-06-10
Section custom spacing

gilbox:
Could use some advice for adding a test...


Listable/Sources/Section.swift
```diff
@@ -152,10 +152,13 @@ public extension Section
     struct Layout : Equatable
     {
         public var width : CustomWidth
+
+        public var customSpacing : CGFloat?
```
kyleve, line 156: Maybe `interSectionSpacingWithFooter`?

kyleve:
For testing, I have a temp test kicking around here: https://github.com/kyleve/Listable/blob/main/Listable/Tests/Layout/List/DefaultListLayoutTests.swift#L71 which I'm using to allow for refactoring of the layout system without breaking things (after which will backfill more tests). Could update that with this property to test it does what's expected

PR #193, opened 2020-08-11
Support tappable headers

kyleve:
This PR introduces support for tappable headers and footers, by adding an `onTap` handler to `HeaderFooter`. When provided, the header becomes tappable, and you can provide a `pressedBackgroundView` to appear, similar to the `selectedBackgroundView` of cells.

There's some unrelated changes to how zIndexing is done to improve determinism in layout, and there are also some fixes to bugs I found alongside this work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 - [Simplify `Sizing` now that enums support default associated values.](https://github.com/kyleve/Listable/pull/189). Now instead of separate `.thatFits` and `.thatFitsWith(Constraint)` enums, there is a single `.thatFits(Constraint = .noConstraint)` case (the same applies for `autolayout`).
 
-- Changed [how `zIndexes` are assigned to header and items, and support tapping headers / footers](https://github.com/kyleve/Listable/pull/193). This allows registering an `onTap` handler for any HeaderFooter, and providing a backgrouned to display while the tap's press is active.
+- Changed [how `zIndexes` are assigned to header and items, and support tapping headers / footers](https://github.com/kyleve/Listable/pull/193). This allows registering an `onTap` handler for any HeaderFooter, and providing a background to display while the tap's press is active.
 
 ### Misc
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 - [Simplify `Sizing` now that enums support default associated values.](https://github.com/kyleve/Listable/pull/189). Now instead of separate `.thatFits` and `.thatFitsWith(Constraint)` enums, there is a single `.thatFits(Constraint = .noConstraint)` case (the same applies for `autolayout`).
 
+- Changed [how `zIndexes` are assigned to header and items, and support tapping headers / footers](https://github.com/kyleve/Listable/pull/193). This allows registering an `onTap` handler for any HeaderFooter, and providing a backgrouned to display while the tap's press is active.
+
 ### Misc
 
 # Past Releases
--- a/Listable/Sources/Internal/HeaderFooterContentView.swift
+++ b/Listable/Sources/Internal/HeaderFooterContentView.swift
@@ -100,7 +100,10 @@ final class HeaderFooterContentView<Content:HeaderFooterContent> : UIView
         
         switch state {
 
-        case .possible, .began, .changed:
+        case .possible:
+            break
+        
+        case .began, .changed:
             self.pressedBackground.alpha = 1.0
             
         case .ended, .cancelled, .failed:
--- a/Listable/Sources/Internal/HeaderFooterContentView.swift
+++ b/Listable/Sources/Internal/HeaderFooterContentView.swift
@@ -81,14 +81,14 @@ final class HeaderFooterContentView<Content:HeaderFooterContent> : UIView
         self.removeGestureRecognizer(self.pressRecognizer)
         
         if self.onTap != nil {
-            self.accessibilityTraits = [.button]
+            self.accessibilityTraits = [.header, .button]
             
             self.pressedBackground.isHidden = false
             self.pressedBackground.alpha = 0.0
             
             self.addGestureRecognizer(self.pressRecognizer)
         } else {
-            self.accessibilityTraits = []
+            self.accessibilityTraits = [.header]
             
             self.pressedBackground.isHidden = true
         }
--- a/Listable/Sources/Internal/HeaderFooterContentView.swift
+++ b/Listable/Sources/Internal/HeaderFooterContentView.swift
@@ -1,0 +1,166 @@
+//
+//  HeaderFooterContentView.swift
+//  Listable
+//
+//  Created by Kyle Van Essen on 8/11/20.
+//
+
+import UIKit
+
+
+final class HeaderFooterContentView<Content:HeaderFooterContent> : UIView
+{
+    //
+    // MARK: Properties
+    //
+    
+    typealias OnTap = () -> ()
+    
+    var onTap : OnTap? = nil {
+        didSet { self.updateIsTappable() }
+    }
+    
+    let content : Content.ContentView
+    let background : Content.BackgroundView
+    let pressedBackground : Content.PressedBackgroundView
+    
+    private let pressRecognizer : PressGestureRecognizer
+    
+    //
+    // MARK: Initialization
+    //
+    
+    override init(frame: CGRect) {
+        
+        let bounds = CGRect(origin: .zero, size: frame.size)
+        
+        self.content = Content.createReusableContentView(frame: bounds)
+        self.background = Content.createReusableBackgroundView(frame: bounds)
+        self.pressedBackground = Content.createReusablePressedBackgroundView(frame: bounds)
+        
+        self.pressRecognizer = PressGestureRecognizer()
+        self.pressRecognizer.minimumPressDuration = 0.0
+        self.pressRecognizer.allowableMovementAfterBegin = 5.0
+        
+        super.init(frame: frame)
+        
+        self.pressRecognizer.addTarget(self, action: #selector(pressStateChanged))
+     
+        self.addSubview(self.background)
+        self.addSubview(self.pressedBackground)
+        self.addSubview(self.content)
+        
+        self.updateIsTappable()
+    }
+    
+    @available(*, unavailable)
+    required init?(coder: NSCoder) { fatalError() }
+    
+    //
+    // MARK: UIView
+    //
+    
+    override func sizeThatFits(_ size: CGSize) -> CGSize {
+        self.content.sizeThatFits(size)
+    }
+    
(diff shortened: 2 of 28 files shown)


kyleve:
We should allow pinning items, so these headers can be pinned

Listable/Sources/Layout/ListLayout/ListLayout.swift
```diff
@@ -58,6 +58,43 @@ public protocol AnyListLayout : AnyObject
...
+        // list header        2
+        //
+        // section header     1
+        // item               0
+        // item               0
+        // section footer    -1
+        //
+        // section header     2
+        // item               1
+        // section footer     0
```
kylebshr, line 79: I always thought of footers on the same z-index as headers, or at least above items but below headers (sticky footers are a thing right?)

kyleve: There's no sticky footers right now; but I suppose that could be a thing.

Tbh, I need to re-think this PR more; im not super happy with it.

Demo/Sources/Demos/Demo Screens/AccordianViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+//  Created by Kyle Van Essen on 8/10/20.
+//  Copyright © 2020 Kyle Van Essen. All rights reserved.
+//
+
+
+import BlueprintLists
+import BlueprintUICommonControls
+
+
+final class AccordianViewController : ListViewController
```
lokae0, line 14: typo- Accordion

kyleve: TIL I have been spelling this wrong my entire life, apparently

kylebshr: I definitely thought the "a" looked correct.. it's the berenstain bears all over again

Listable/Sources/Internal/HeaderFooterContentView.swift
```diff
@@ -0,0 +1,156 @@
...
+    }
+    
+    override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent) {
+        super.touchesBegan(touches, with: event)
+        
+        self.initialPoint = self.location(in: self.view)
+    }
+    
+    override func canPrevent(_ preventedGestureRecognizer: UIGestureRecognizer) -> Bool {
+        false
```
kyleve, line 140: This should probably compare against the gesture recognizer on the list view itself? Not just any arbitrary gesture?

Listable/Sources/Layout/ListLayout/ListLayout.swift
kyleve: This has gone back to being flat based on element type, since making headers tappable means you dont need to order these as intricately.

Demo/Sources/Demos/Demo Screens/AccordianViewController.swift
kyleve: Fixed

Listable/Sources/Layout/SupplementaryKind.swift
```diff
@@ -8,7 +8,7 @@
 import Foundation
 
 
-enum SupplementaryKind : String, CaseIterable
+public enum SupplementaryKind : String, CaseIterable
```
kyleve, line 11: Doesnt need to be public anymore

Listable/Sources/ListView/ListView.VisibleContent.swift
```diff
@@ -10,12 +10,12 @@ import Foundation
 
 extension ListView
 {
-    struct VisibleContent
+    final class VisibleContent
```
kyleve, line 13: This is needed to fix an unrelated bug that I found working on this; needs to be a class to avoid exclusive access crashes.

kyleve:
Alright, this should be good to go now

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 - [Simplify `Sizing` now that enums support default associated values.](https://github.com/kyleve/Listable/pull/189). Now instead of separate `.thatFits` and `.thatFitsWith(Constraint)` enums, there is a single `.thatFits(Constraint = .noConstraint)` case (the same applies for `autolayout`).
 
+- Changed [how `zIndexes` are assigned to header and items, and support tapping headers / footers](https://github.com/kyleve/Listable/pull/193). This allows registering an `onTap` handler for any HeaderFooter, and providing a backgrouned to display while the tap's press is active.
```
kylebshr, line 15:
```suggestion
- Changed [how `zIndexes` are assigned to header and items, and support tapping headers / footers](https://github.com/kyleve/Listable/pull/193). This allows registering an `onTap` handler for any HeaderFooter, and providing a background to display while the tap's press is active.
```

Listable/Sources/Internal/HeaderFooterContentView.swift
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+    
+    @objc private func pressStateChanged() {
+        
+        let state = self.pressRecognizer.state
+        
+        switch state {
+
+        case .possible, .began, .changed:
```
kylebshr, line 103: Is it right to include `possible` here? The docs say that's the "default" state, but maybe that's not true for concrete subclasses?

Listable/Sources/Internal/HeaderFooterContentView.swift
```diff
@@ -0,0 +1,166 @@
...
+            self.pressedBackground.alpha = 1.0
+            
+        case .ended, .cancelled, .failed:
+            let didEnd = state == .ended
+            
+            UIView.animate(withDuration: didEnd ? 0.1 : 0.0) {
```
kylebshr, line 109: `UITableView` doesn't animate deselection except when specified, and it's more like 0.3s - not sure we shouldn't do this, but seems rather arbitrary

Listable/Sources/Internal/HeaderFooterContentView.swift
```diff
@@ -0,0 +1,166 @@
...
+                self.onTap?()
+            }
+            
+        @unknown default: break
+        }
+    }
+}
+
+
+fileprivate final class PressGestureRecognizer : UILongPressGestureRecognizer {
```
kylebshr, line 123: I'm curious why you're using a long press gesture for this? Does it recognize a quick tap?

Listable/Sources/Layout/CollectionViewLayout.swift
```diff
@@ -347,12 +347,23 @@ final class CollectionViewLayout : UICollectionViewLayout
...
-        
+                
         self.viewProperties = CollectionViewLayoutProperties(collectionView: view)
     }
     
+    private func performLayoutUpdate()
+    {
+        let view = self.collectionView!
+        
+        self.layout.positionStickySectionHeadersIfNeeded(in: view)
```
kylebshr, line 362: Curious why you moved this out of `updateLayout` - is there a case where you'd update the layout with positioning the headers?

Listable/Sources/Layout/ListLayout/ListLayoutContent.swift
```diff
@@ -303,15 +305,19 @@ public extension ListLayoutContent
...
+            )
         }
         
         let kind : SupplementaryKind
         let layout : HeaderFooterLayout
         let measurer : (Sizing.MeasureInfo) -> CGSize
                 
         let isPopulated : Bool
-                
+
```
kylebshr, line 320: Doesn't actually matter ofc, but there's a checkbox in Xcode to strip whitespace from empty lines to prevent this in diffs :)

Listable/Sources/Internal/HeaderFooterContentView.swift
kyleve: I wanted this animated, since it's basically like a button press, and it looks nicer when the press ends with an animation, IMO.

Listable/Sources/Internal/HeaderFooterContentView.swift
kyleve: I needed a continuous gesture recognizer, a tap recognizer is discrete (it calls you on tap and then is done), which isnt how row selection works in list views; eg how you can press down, hold, and then release to trigger a row selection.

Or put another way, this mimics the row selection behavior UICollectionView provides.

Listable/Sources/Layout/ListLayout/ListLayoutContent.swift
kyleve: Huh, TIL!

Listable/Sources/Layout/CollectionViewLayout.swift
kyleve: Since this was getting called in multiple layouts, I figured it made sense to centralize it here; which seemed to make sense.

Listable/Sources/Internal/HeaderFooterContentView.swift
kyleve: (And it does recognize a tap, since I set the min duration to 0:)

```
        self.pressRecognizer = PressGestureRecognizer()
        self.pressRecognizer.minimumPressDuration = 0.0
        self.pressRecognizer.allowableMovementAfterBegin = 5.0
```

kylebshr: Ah gotcha, just noticed the `minimumPressDuration` set to 0 above - missed that before

Listable/Sources/Internal/HeaderFooterContentView.swift
kyleve: Ahh yeah, doh. I wonder how this works at all...

kyleve: Fixed!

Listable/Sources/Internal/HeaderFooterContentView.swift
```diff
@@ -0,0 +1,169 @@
...
+        
+        if self.onTap != nil {
+            self.accessibilityTraits = [.button]
+            
+            self.pressedBackground.isHidden = false
+            self.pressedBackground.alpha = 0.0
+            
+            self.addGestureRecognizer(self.pressRecognizer)
+        } else {
+            self.accessibilityTraits = []
```
kyleve, line 91: This should have `header` in it.

Listable/Sources/Internal/Presentation State/PresentationState.HeaderFooterState.swift
```diff
@@ -95,16 +95,24 @@ extension PresentationState
             cache.push(view, with: self.model.reuseIdentifier)
         }
         
-        func createReusableHeaderFooterView(frame : CGRect) -> UIView
```
kyleve, line 98: Unused

PR #197, opened 2020-08-20
Switch from isEmpty to contains(:), in order to make it clearer what you're asking if the list contains.

kyleve:


Listable/Sources/Content.swift
```diff
@@ -32,8 +32,35 @@ public struct Content
         return self.sections.reduce(0, { $0 + $1.items.count })
     }
     
-    public var isEmpty : Bool {
-        return self.sections.isEmpty || self.sections.allSatisfy { $0.isEmpty }
+    public func contains(any filters : Set<IsEmptyFilters> = [.items]) -> Bool {
```
kyleve, line 35: Test me

Listable/Sources/Section.swift
```diff
@@ -22,8 +22,29 @@ public struct Section
     
     public var items : [AnyItem]
     
-    public var isEmpty : Bool {
-        return self.items.isEmpty
+    public func contains(any filters : Set<Content.IsEmptyFilters> = [.items]) -> Bool {
```
kyleve, line 25: Test me

Listable/Sources/Content.swift
```diff
@@ -224,6 +265,43 @@ public struct Content
 }
 
 
+public extension Content {
+    
+    /// A filter enum which allows you to query the types of content contained in a `Content` or `Section` object.
+    enum IsEmptyFilters : Hashable, CaseIterable {
```
lokae0, line 271: How do you feel about naming this `ContentFilters` or something similar? I get the intent behind `IsEmptyFilters`, but it's kinda hard to both read and understand

PR #199, opened 2020-08-24
Allow begin observing the keyboard notifications without the view being loaded

kyleve:
So this is kinda gross, but, no other great solution.

Problem: We receive keyboard frames based on system NSNotifcations. If the keyboard is already visible when a list is created, we previously had no way of knowing what the current keyboard frame was, and thus, we had no way of knowing how to adjust the contentInset of the list view to avoid keyboard overlap.

Solution: On app startup, register for keyboard notifications, and use a single shared observer.


Listable/Sources/Internal/KeyboardObserver.swift
```diff
@@ -42,9 +42,25 @@ protocol KeyboardObserverDelegate : AnyObject {
  */
 final class KeyboardObserver {
     
+    static let shared : KeyboardObserver = KeyboardObserver()
+    
     private let center : NotificationCenter
     
-    weak var delegate : KeyboardObserverDelegate?
+    private var delegates : [Delegate] = []
```
kyleve, line 49: Need to GC these

kylebshr:
Any reason not to use a module initializer (like the second answer [here](https://stackoverflow.com/questions/49029515/is-there-an-alternative-to-initialize-in-macos-now-that-swift-has-deprecated-i)) and set this up automatically?

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
```diff
@@ -125,11 +175,13 @@ final class KeyboardObserver {
          */
         let animationOptions = UIView.AnimationOptions(rawValue: new.animationCurve << 16)
         
-        self.delegate?.keyboardFrameWillChange(
-            for: self,
-            animationDuration: new.animationDuration,
-            options: animationOptions
-        )
+        self.delegates.forEach {
```
kyleve, line 178: Hold a local ref to this; since this could result in something making a _new_ list, it might violate Swift's unique access on structs constraint.

kyleve:
@kylebshr: Do you mean the `static void __attribute__ ((constructor)) Initer()` bit?

kylebshr:
Yep - won't that get called when the module is loaded? (disclaimer I have no idea what I'm talking about)

kyleve:
@kylebshr So the C-style constructors should generally be avoided in this kind of case; since this is calling across ObjC types and across frameworks. If you're doing purely local, C-only setup, the constructor call would work, but in this case, we need to wait until the runtime gives us a chance to set things up.

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
```diff
@@ -42,21 +42,42 @@ protocol KeyboardObserverDelegate : AnyObject {
...
+    static let shared : KeyboardObserver = KeyboardObserver(center: .default)
+    
+    /// Allow logging to the console if app startup-timed shared instance startup did not
+    /// occur; this could cause bugs for the reasons outlined above.
+    fileprivate static var didSetupSharedInstanceDuringAppStartup = false
+    
     private let center : NotificationCenter
     
-    weak var delegate : KeyboardObserverDelegate?
+    var delegates : [Delegate] = []
```
kyleve, line 62: `private(set) var`

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
```diff
@@ -42,21 +42,42 @@ protocol KeyboardObserverDelegate : AnyObject {
...
+    
+    struct Delegate {
+        weak var value : KeyboardObserverDelegate?
```
kyleve, line 65: `private(set) weak var`

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
```diff
@@ -67,6 +88,35 @@ final class KeyboardObserver {
...
+    
+    func remove(delegate : KeyboardObserverDelegate) {
+        self.delegates.removeAll {
+            $0.value === delegate
+        }
+        
+        self.removeDeallocatedDelegates()
+    }
+    
+    func removeDeallocatedDelegates() {
```
kyleve, line 114: `private func`

kyleve:
@kylebshr: Mind giving this a look today? I'd like to merge today to unblock Jason

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
```diff
@@ -125,11 +175,19 @@ final class KeyboardObserver {
...
-        self.delegate?.keyboardFrameWillChange(
-            for: self,
-            animationDuration: new.animationDuration,
-            options: animationOptions
-        )
+        /// Make a local copy of the array. Because we're calling out to external consumers,
+        /// it's possible (but unlikely) that changing the keyboard visibility on a list could cause
+        /// _another_ list to be allocated, which would mutate the `self.delegates` array
+        /// while we are mutating it, violating Swift's exclusive access clause.
+        let delegates = self.delegates
```
kylebshr, line 182: I don't know a ton about exclusive access, but isn't this a non-issue because all these changes would happen on a main thread? I don't see how there are conflicting accesses

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
```diff
@@ -190,3 +248,38 @@ extension KeyboardObserver
         }
     }
 }
+
+
+extension KeyboardObserver {
+    
+    /// Called by `ListView` on setup, to warn developers
+    /// if something has gone wrong with keyboard setup.
+    static func logKeyboardSetupWarningIfNeeded() {
```
kylebshr, line 257: Do you think the logging is still needed now that set up is automatic? Is there a case where this wouldn't get set up?

kylebshr:
@kyleve totally looked last night and forgot to approve, had some questions but no need to block

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
kyleve: Since I'm doing something that's a little clever to get the registration working, I wanted to add an obvious log statement if it stops working for whatever reason.

kylebshr: Reasonable 👍

Listable/Sources/Internal/KeyboardObserver/KeyboardObserver.swift
kyleve: It can happen due to nested accesses as well, eg:

```
self.things.forEach { thing in
    self.things.append(newThing)
}
```

Though after testing it, this no longer happens for arrays; I think they're smart about arrays now, likely since they're COW internally (but this was an issue with the original exclusive access implementation). Going to remove this bit!

https://github.com/apple/swift-evolution/blob/master/proposals/0176-enforce-exclusive-access-to-memory.md

kyleve: (I ran into this last week with the `VisibleContent` struct, which is why I did this here, but yeah, seems like arrays are somehow immune, as are im guessing all the stdlib collections.)

PR #202, opened 2020-09-01
Allow setting the sizing type on List

kyleve:


Listable/Sources/ListView/ListView.swift
```diff
@@ -488,10 +488,57 @@ public final class ListView : UIView
...
+        
+        view.frame.size = fittingSize
+        
+        /// Push the updated content into the list.
+        
+        view.configure(with: properties)
+        
+        let size = view.contentSize
+        
+        /// Now that we have the the measured contentSize,
```
kyleve, line 515: Do we actually need to do this? Eg, if every closure is correctly weak, it should be fine. Though we might want to make a special case for this underlying prototype list.

Listable/Sources/Layout/ListLayout/LayoutDescription.swift
```diff
@@ -78,14 +78,18 @@ extension LayoutDescription
         
         // MARK: AnyLayoutDescriptionConfiguration
         
-        public func createEmptyLayout() -> AnyListLayout
+        func createEmptyLayout(
```
kyleve, line 81: Changing this signature makes the new `ListSizing` tests pass; before not passing these through caused a delay in correct layout by one loop because we had to wait for a rebuild of the layout.

Listable/Sources/ListView/ListSizing.swift
```diff
@@ -0,0 +1,181 @@
...
+            width: fittingSize.width > 0 ? min(fittingSize.width, size.width) : size.width,
+            height: fittingSize.height > 0 ? min(fittingSize.height, size.height) : size.height
+        )
+    }
+}
+
+
+fileprivate extension ListProperties
+{
+    func toListSizingSizable() -> ListProperties {
```
kyleve, line 120: TODO: Decide if I want to do this, or if I want to just mark an entire list as a sizing list. Likely leaning towards the second... I think?

Listable/Sources/HeaderFooter/HeaderFooter.swift
```diff
@@ -72,6 +72,27 @@ public struct HeaderFooter<Content:HeaderFooterContent> : AnyHeaderFooter
         self.reuseIdentifier = ReuseIdentifier.identifier(for: Content.self)
     }
     
+    /// Internal init with all parameters, none with default values.
```
kyleve: Remove

Listable/Sources/Internal/Presentation State/PresentationState.swift
```diff
@@ -21,6 +21,12 @@ final class PresentationState
     var overscrollFooter : HeaderFooterViewStatePair = .init()
     
     var sections : [PresentationState.SectionState]
+    
+    var performsContentCallbacks : Bool = true {
+        didSet {
```
kyleve: Remove

PR #206, opened 2020-09-11
Create a blueprint backed item or header without needing to define your own ItemContent

kyleve:
A common ask is for a way to display rows without needing to create entirely new `BlueprintItemContent` instances without needing to define a whole new type. This PR lets you do this for headers and footers:

```
section += podcasts.map { podcast in
    
    Item(podcast, identifier: \.name) { _, _ in
        PodcastElement(podcast: podcast)
    } background: { _, _ in
        Box(backgroundColor: .white, cornerStyle: .square)
    } selectedBackground: { _, _ in
        Box(backgroundColor: .lightGray, cornerStyle: .square)
    }
}
```


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
```
kyleve, line 1: Intentional changes in here; no reason to pin these during `bundle update`.

Listable/Sources/Item/Item.swift
```diff
@@ -77,6 +77,9 @@ public struct Item<Content:ItemContent> : AnyItem
     
     public typealias Build = (inout Item) -> ()
```
kyleve, line 78: Document me

kylebshr (approved):
LGTM, sorry this went unnoticed for so long

ListableUI/Sources/Item/Item.swift
```diff
@@ -81,6 +81,9 @@ public struct Item<Content:ItemContent> : AnyItem
```
kyleve, line 81: Revert these changes

kyleve:
Thanks! No rush, not an urgent one.

PR #207, opened 2020-09-17
Allow `ItemCell.ContentViewContainer` subviews to specify accessibility params

lokae0:
Don't set `isAccessibilityElement = true` for `ItemCell.ContentViewContainer`.

This fixes a couple of related bugs: 
* The content container's subviews could not specify accessibility parameters (e.g. labels, values, etc).
* Elements within the container could not be interacted with via Voiceover (e.g. swipe buttons).

Also:
* Add accessibility labels to demo app for testing
* Resolve a warning where the selector passed to `UIAccessibilityCustomAction.init` should return a bool
* Add a unit test

I tested on-device with Voiceover Rotor actions menu. Ensured that swipe actions and custom accessibility actions are still working as intended.

Demo:
https://youtu.be/dTTENavTYpc

### DIFF
--- a/Listable/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/Listable/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -31,8 +31,6 @@ extension ItemCell {
 
             super.init(frame: frame)
 
-            isAccessibilityElement = true
-
             self.addSubview(self.contentView)
 
             NotificationCenter.default.addObserver(
@@ -223,8 +221,9 @@ extension ItemCell {
             }
         }
 
-        @objc private func performAccessibilityAction(_ action: AccessibilitySwipeAction) {
+        @objc private func performAccessibilityAction(_ action: AccessibilitySwipeAction) -> Bool {
             action.action.handler(self.didPerformAction)
+            return true
         }
 
         private func configureAccessibilityActions(for actions: [SwipeAction]) {
--- a/Demo/Sources/Demos/Demo Screens/CollectionViewAppearance.swift
+++ b/Demo/Sources/Demos/Demo Screens/CollectionViewAppearance.swift
@@ -98,6 +98,7 @@ struct DemoItem : BlueprintItemContent, Equatable
             $0.color = info.state.isActive ? .white : .darkGray
         }
         .inset(horizontal: 15.0, vertical: 10.0)
+        .accessibility(label: self.text, traits: [.button])
     }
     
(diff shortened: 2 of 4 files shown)


Listable/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -223,8 +221,9 @@ extension ItemCell {
             }
         }
 
-        @objc private func performAccessibilityAction(_ action: AccessibilitySwipeAction) {
+        @objc private func performAccessibilityAction(_ action: AccessibilitySwipeAction) -> Bool {
```
kyleve, line 224: Needed?

kyleve (approved):
Thanks!

Listable/Sources/Internal/ItemCell.ContentViewContainer.swift
lokae0: Added this to address a debug console warning. Documentation for `UIAccessibilityCustomAction.init` says:

```
The method signature must take one of the following forms:

- (BOOL)myPerformActionMethod
- (BOOL)myPerformActionMethod:(UIAccessibilityCustomAction *)action
```

The bool indicates action fail / success, although naturally this isn't documented at all ¯\\\_(ツ)\_/¯

kyleve: Ahh interesting! TIL

kylebshr:
I think the reason I made it an accessibility element in the first place was to use `accessibilityCustomActions` for accessing the swipe actions through voice over. Can you verify that this doesn't break that?

lokae0:
> I think the reason I made it an accessibility element in the first place was to use `accessibilityCustomActions` for accessing the swipe actions through voice over (those are what's calling `performAccessibilityAction`). Can you verify that this doesn't break that?

Yep, I tested these thoroughly to make sure they didn't break. You can see & hear this towards the end of the demo video: https://youtu.be/dTTENavTYpc?t=13

When VO reads out "delete" & "save", those are the custom actions being accessed by swiping up & down via Rotor. Then I double tap to perform the save / unsave action on `Item #5`.

kylebshr:
I feel terrible, you even mentioned the video in your messages and I forgot to check it out 😅 thank you so much!

kylebshr (approved):

lokae0:
@kylebshr no worries at all! Thanks for the review

PR #209, opened 2020-09-24
Deprecate iOS 10

kyleve:
See ya!

Also moves us to test on 11/12/13/14


.travis.yml
```diff
@@ -3,12 +3,7 @@ git:
 
 language: swift
 os: osx
-osx_image: xcode11.5
-
-# https://docs.travis-ci.com/user/caching/
-cache:
```
kyleve, line 9: Removing this since I don't think it does a ton now with the Cocoapods CDN... verifying.

kyleve: Ok I think it made it slower, bringing it back

kyleve: It is back

kylebshr (approved):
Hmm I really like how all the tests are named with this set up - that's one downside of just using a matrix like I did with BP

kyleve:
Tbh, one reason I did it this way since i couldn't figure out how to get travis' matrices working. Other reason I did it this way is so that you can locally run the `run_all_tests.sh` script to run the tests across all iOS versions easily; useful for snapshot tests.

kylebshr:
Super useful. Maybe we can steal it some day

PR #212, opened 2020-09-25
Allow publishing to CocoaPods

kyleve:
This fixes various errors that broke publishing to cocoapods; which we can now do! Because the tests depend on local pods; I've wrapped those declarations in a check, and included instructions in the releasing.md, which is just a copy of blueprint's releasing info.

### DIFF
--- a/BlueprintUILists.podspec
+++ b/BlueprintUILists.podspec
@@ -1,12 +1,12 @@
 
 Pod::Spec.new do |s|
   s.name         = 'BlueprintUILists'
-  s.version      = '0.9.0'
+  s.version      = '0.10.0'
   s.summary      = 'Declarative list views for iOS apps that deploy back to iOS 11.0.'
   s.homepage     = 'https://github.com/kyleve/Listable'
-  s.license      = { type: 'Proprietary', text: "© 2020 Square, Inc." }
-  s.author       = { 'iOS Team' => 'seller-ios@squareup.com' }
-  s.source       = { git: 'Not Published', tag: "podify/#{s.version}" }
+  s.license      = 'Apache License, Version 2.0'
+  s.author       = { 'Kyle' => 'k@squareup.com' }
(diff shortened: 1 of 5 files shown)


BlueprintUILists.podspec
```diff
@@ -1,12 +1,12 @@
...
   s.name         = 'BlueprintUILists'
-  s.version      = '0.9.0'
+  s.version      = '0.10.0'
   s.summary      = 'Declarative list views for iOS apps that deploy back to iOS 11.0.'
   s.homepage     = 'https://github.com/kyleve/Listable'
-  s.license      = { type: 'Proprietary', text: "© 2020 Square, Inc." }
-  s.author       = { 'iOS Team' => 'seller-ios@squareup.com' }
-  s.source       = { git: 'Not Published', tag: "podify/#{s.version}" }
+  s.license      = 'Apache License, Version 2.0'
+  s.author       = { 'Kyle' => 'k@squareup.com' }
```
kyleve, line 8: Not sure what else to put here, since our internal email lists aren't publicly emailable...

CHANGELOG.md
```diff
@@ -187,12 +197,16 @@
...
 - Added support for [conditionally scrolling to items](https://github.com/kyleve/Listable/pull/129) on insert, based on the `shouldPerform` block passed to the `AutoScrollAction`.
 
-## Earlier
+## Earlier - Internal Only
 
 Earlier releases were ad-hoc and not tracked. To see all changes, please reference [closed PRs on Github](https://github.com/kyleve/Listable/pulls?q=is%3Apr+is%3Aclosed).
+
+
+[Main]: https://github.com/square/Blueprint/compare/0.10.0...HEAD
+[0.10.0]: https://github.com/square/Blueprint/compare/0.9.0...0.10.0
```
kyleve, line 212: Note: I published 0.9.0 as a "dummy" version to let us have this example in the changelog for future cargo culting, however its the same as 0.9.0; since because before today versioning was pretty random.

PR #222, opened 2020-10-22
Cleanup scroll insets stuff

kyleve:
Fixed up some incorrect code from a very early version of Listable: Rename scrollInsets to scrollIndicatorInsets since that's what it was really meant for; and then clean up the inset handling within ListView itself to centralize it in one place.


ListableUI/Sources/ListView/ListSizing.swift
```diff
@@ -99,7 +99,7 @@ extension ListView
             horizontal: CGSize(width: 100.0, height: fittingSize.height)
         )
         
-        let size = view.contentSize
+        let size = properties.scrollInsets.appliedTo(contentSize: view.contentSize)
```
kyleve, line 102: Revert

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -275,19 +275,84 @@ public final class ListView : UIView
...
         )
-
-        self.collectionView.contentInset = insets
-        self.collectionView.scrollIndicatorInsets = insets
+        
+        let scrollInsets = Init(self.scrollInsets.toUIEdgeInsets) {
+            $0.bottom = bottomInset
+        }
+        
+        let contentInsets = UIEdgeInsets(
```
kyleve, line 327: This should be 0 except for bottom

ListableUI/Sources/ScrollIndicatorInsets.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  ScrollIndicatorInsets.swift
+//  ListableUI
+//
+//  Created by Kyle Van Essen on 11/13/19.
+//
+
+
+public struct ScrollIndicatorInsets : Equatable
```
kyleve, line 9: This type is nice since it gives you a better init for UIEdgeInsets, but I guess I can just add a public init to UIEI instead.

kyleve: Removed

PR #225, opened 2020-10-27
Pipe Blueprint's environment through during layout and measurement by creating an environment of our own

kyleve:
This PR updates `List` to ensure that we pass `BlueprintUI.Environment` through to the inner blueprint views within each item and header/footer within the list. This is done by writing `BlueprintUI.Environment` into our own `ListEnvironment` – which is now piped through to each item and header/footer within the list, and then, we reach this into environment and pull out the bp environment within `BlueprintItemContent` and `BlueprintHeaderFooterContent` to continue to pass the values down.


ListableUI/Sources/ListView/ListView.DataSource.swift
```diff
@@ -11,6 +11,7 @@ internal extension ListView
     final class DataSource : NSObject, UICollectionViewDataSource
     {
         unowned var presentationState : PresentationState!
+        var environment : ListEnvironment = .empty
```
kyleve, line 14: unowned var too? Or at least an IUO?

BlueprintUILists/Sources/BlueprintItemContent.swift
```diff
@@ -147,4 +147,14 @@ public extension BlueprintItemContent
         
         return view
     }
+    
+    private func wrap(element : Element?, inBlueprintEnvironmentFrom environment : ListEnvironment) -> Element? {
```
kyleve, line 151: Put into an internal extension to share w/ headerfooter version

BlueprintUILists/Sources/List.swift
```diff
@@ -76,27 +76,82 @@ public struct List : Element
...
+                config.apply { listView in
+                    listView.configure(with: self.properties)
+                }
+            }
+        }
+    }
+}
+
+
+internal extension ListEnvironment {
```
kyleve, line 144: Remove `internal`

BlueprintUILists/Sources/List.swift
```diff
@@ -76,27 +76,82 @@ public struct List : Element
...
+    
+    var blueprintEnvironment : Environment {
+        get { self[BlueprintEnvironmentListKey.self] }
+        set { self[BlueprintEnvironmentListKey.self] = newValue }
+    }
+}
+
+internal enum BlueprintEnvironmentListKey : ListEnvironmentKey {
```
kyleve, line 152: Remove `internal`

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 - [Introduce `onSelectionChanged` on `ListStateObserver`](https://github.com/kyleve/Listable/pull/223) to allow observing when the selected rows change.
 
+- [Pass through `BlueprintUI.Environment` to the `Element`s being rendered from `BlueprintItemContent` and `BlueprintHeaderFooterContent`](https://github.com/kyleve/Listable/pull/225). This ensures that the content you put into a `List` respect the `BlueprintUI.Environment` of the `List` itself. This PR also introduces `ListEnvironment` to facilitate this, which allows similar passthrough of environment variables within Listable.
```
kyleve, line 11: `will respect`...

BlueprintUILists/Sources/BlueprintHeaderFooterContent.swift
```diff
@@ -139,4 +142,14 @@ public extension BlueprintHeaderFooterContent
         
         return view
     }
+    
+    private func wrap(element : Element?, inBlueprintEnvironmentFrom environment : ListEnvironment) -> Element? {
```
kyleve, line 146: Put into an internal extension to share w/ headerfooter version

kylebshr (commented):
High level thoughts/questions:

- Do we need the env to also exist/get passed through in the `ListableUI` types? If you think it's useful I'm fine with it, but I'm not sure I see the utility of it?
- If the answer to ^ is no, we don't need it, I don't think we should add a `ListEnvironment` type at all, and just pipe through the BP environment. We can add keys and use the BP environment to pass stuff down in `BlueprintUILists`

kyleve:
I guess I'm not seeing how we'd get the Blueprint `Environment` down to each header and footer without some sort of environment-like thing in Listable – the data flow is:

1) You make a `List`
2) It goes into an outer `BlueprintView`
3) Blueprint updates the inner `ListView`
4) The inner `ListView` then renders cells by creating their inner `BlueprintView`s
5) Those inner views need the outer view's environment when they're getting rendered – their lifetime is not known by the outer `List`.

The key part being there's an inner extra layer of blueprint views whose lifetime is not known or controlled by users of the `List` itself.

kyleve:
(I _do_ wonder if there's some way that we can / need to solve this more generally in blueprint at some point – eg this will _also_ break for any nested BP views in Market itself; which is not good for when we let people inject custom content, or we want to leverage the environment ourselves... But that's a whole thing)

BlueprintUILists/Tests/ListTests.swift
```diff
@@ -0,0 +1,135 @@
...
+            list("section") { section in
+                
+                section.header = HeaderFooter(TestHeaderContent(wasCalled: callback))
+                section.footer = HeaderFooter(TestHeaderContent(wasCalled: callback))
+                
+                section += TestItemContent(wasCalled: callback)
+                section += TestItemContent(wasCalled: callback)
+            }
+        }.adaptedEnvironment { env in
+            env[TestingKey.self] = true
```
zadr, line 39: worth an `XCTAssertTrue(env[TestingKey.self])` somewhere at the end of the test (similar to `TestHeaderContent` and `TestItemContent`)?

ListableUI/Sources/Internal/SupplementaryContainerView.swift
```diff
@@ -84,14 +86,19 @@ final class SupplementaryContainerView : UICollectionReusableView
...
+                    frame: self.bounds,
+                    environment: self.environment
+                )
             } else {
                 self.content = nil
             }
         }
     }
     
-    var reuseCache : ReusableViewCache?
```
zadr, line 94: when would the reuseCache be nil— if we init something instead of dequeuing it? 

if so, something like `precondition(reuseCache != nil, "must dequeue a Supplementary Container View")` might make it clearer. definitely a nit, though

ListableUI/Sources/ListEnvironment.swift
```diff
@@ -0,0 +1,80 @@
...
+///     to views : ItemContentViews<Self>,
+///     for reason: ApplyReason,
+///     with info : ApplyItemContentInfo
+/// ) {
+///    switch info.environment.myLayoutType {
+///       ...
+///    }
+/// }
+/// ```
+public struct ListEnvironment {
```
zadr, line 44: maybe something like
```
public typealias ListEnvironment = BlueprintUI.Environment
public typealias ListEnvironmentKey = BlueprintUI.EnvironmentKey
```
for now? so you own the type but don't have to rebuild the world until you have a use-case to do so

zadr (approved):

ListableUI/Sources/ListEnvironment.swift
kyleve: `ListableUI` itself doesnt depend on `BlueprintUI`, so, can't do that.

ListableUI/Sources/Internal/SupplementaryContainerView.swift
kyleve: Because these are `UICollectionReusableView`; the init method is called by UICollectionView – you can't customize it. Thus, using IUOs to make things explode if these properties haven't been injected, without having to unwrap it at every callsite.

zadr:
> (I _do_ wonder if there's some way that we can / need to solve this more generally in blueprint at some point – eg this will _also_ break for any nested BP views in Market itself; which is not good for when we let people inject custom content, or we want to leverage the environment ourselves... But that's a whole thing)

My naive expectation was that `EnvironmentReader` would keep going up some internal tree until it found an environment to use

So if the tree was
```
0. UIView
1. - AdaptedEnvironment
2. -- BlueprintElement
3. --- UIView
4. ---- BlueprintElement
5. ----- EnvironmentReader
6. ------ BlueprintElementtent
```
so instead of the EnvironmentReader at level 5 fizzling out at level 3 because it found a UIView, it would keep traversing the view hierarchy to see if any parent was a view backed by a BlueprintElement and could provide an Environment to use

but on second thought, going up the tree is likely messy implementation (e.g.: would it require the description view to have a back reference to any ElementContent in case it holds an Environment?) and definitely not as nice as explicitly propagating state down the tree

BlueprintUILists/Tests/ListTests.swift
kyleve: Can you clarify what you mean here? I think this would only be testing Blueprint itself; since this is just a normal element.

kyleve:
> > (I _do_ wonder if there's some way that we can / need to solve this more generally in blueprint at some point – eg this will _also_ break for any nested BP views in Market itself; which is not good for when we let people inject custom content, or we want to leverage the environment ourselves... But that's a whole thing)
> 
> My naive expectation was that `EnvironmentReader` would keep going up some internal tree until it found an environment to use
> 
> So if the tree was
> 
> ```
> 0. UIView
> 1. - AdaptedEnvironment
> 2. -- BlueprintElement
> 3. --- UIView
> 4. ---- BlueprintElement
> 5. ----- EnvironmentReader
> 6. ------ BlueprintElementtent
> ```
> 
> so instead of the EnvironmentReader at level 5 fizzling out at level 3 because it found a UIView, it would keep traversing the view hierarchy to see if any parent was a view backed by a BlueprintElement and could provide an Environment to use
> 
> but on second thought, going up the tree is likely messy implementation (e.g.: would it require the description view to have a back reference to any ElementContent in case it holds an Environment?) and definitely not as nice as explicitly propagating state down the tree

Yeah, this – _and_ there's different layout and measurement passes; so you can't just pass it down or "reach up" to find the current status easily.

zadr:
> Yeah, this – _and_ there's different layout and measurement passes; so you can't just pass it down or "reach up" to find the current status easily.

Yeah, that makes a lot of sense, thanks.

I was thinking that the Environment is more independent of the layout/measurement passes so it might be okay. But now that you mention it, I can definitely see people wanting to store per-view information in an environment and propagate it down (e.g. the view size influencing size classes as you get further down the tree is super believable)

BlueprintUILists/Tests/ListTests.swift
zadr: I just missed that it was a plain mock element the first time around and asking from a place of "if we set a state in a test, should we assert that it was set anywhere?"

ListableUI/Sources/Internal/SupplementaryContainerView.swift
kyleve: Added an inline comment to clarify this

kyleve:
@kylebshr any more thoughts here? I can set up some time to run thorough why this is needed (I think?) which might be easier than going back and forth here

kylebshr:
I didn't realize at first that it _had_ to pass through `ListableUI`, so needs it's own environment to not depend on BlueprintUI - makes sense now!

kylebshr (approved):

PR #229, opened 2020-10-29
Add a didPerform callback for AutoScrollAction.

kyleve:
This way actions don't occur in the `shouldPerform` method, which should be idempotent.

### DIFF
--- a/ListableUI/Sources/AutoScrollAction.swift
+++ b/ListableUI/Sources/AutoScrollAction.swift
@@ -30,33 +30,43 @@ public enum AutoScrollAction {
     /// // ID of item which should trigger a scroll event (eg the last item in the list).
     /// let identifier = myItem.identifier
     ///
-    /// let action = .scrollTo(.lastItem, onInsertOf: identifier, position: .init(position: .bottom), animated: true) { state in
+    /// let action = .scrollTo(
+    ///     .lastItem,
+    ///     onInsertOf: identifier,
+    ///     position: .init(position: .bottom),
+    ///     animation: .default
+    /// ) { info in
     ///    // Only scroll to the item if the bottom of the list is already visible.
     ///    return state.isLastItemVisible
+    /// } didPerform : { info in
+    ///     // Called when the scroll action occurs.
     /// }
     /// ```
-    ///
     /// - Parameters
     ///     - destination: Where the list should scroll to on insert. If not specified, the value passed to `onInsertOf` will be used.
     ///     - onInsertOf: The identifier which should trigger the action.
     ///     - position: The position to scroll the list to.
-    ///     - animated: If the scroll action should be animated. Note: Will only animate if the list update itself is animated.
+    ///     - animation: The animation type to perform. Note: Will only animate if the list update itself is animated.
     ///     - shouldPerform: A block which lets you control if the auto scroll action should be performed based on the state of the list.
+    ///     - didPerform: A block which is called when the action is performed. If the item causing insertion is inserted multiple times,
+    ///     this block will be called multiple times.
     ///
     public static func scrollTo(
         _ destination : ScrollDestination? = nil,
         onInsertOf insertedIdentifier: AnyIdentifier,
-        position: ItemScrollPosition,
-        animated: Bool = false,
-        shouldPerform : @escaping (ListScrollPositionInfo) -> Bool = { _ in true }
+        position: ScrollPosition,
+        animation: ScrollAnimation = .none,
+        shouldPerform : @escaping (ListScrollPositionInfo) -> Bool = { _ in true },
+        didPerform : @escaping (ListScrollPositionInfo) -> () = { _ in }
     ) -> AutoScrollAction
     {
         .scrollToItem(onInsertOf: .init(
             destination: destination ?? .item(insertedIdentifier),
             insertedIdentifier: insertedIdentifier,
             position: position,
-            animated: animated,
-            shouldPerform: shouldPerform
+            animation: animation,
+            shouldPerform: shouldPerform,
+            didPerform: didPerform
         ))
     }
 }
@@ -94,17 +104,20 @@ public extension AutoScrollAction
         public var insertedIdentifier : AnyIdentifier
         
         /// The desired scroll position,
-        public var position : ItemScrollPosition
+        public var position : ScrollPosition
         
-        /// Should the change be animated.
+        /// How to animate the change.
         ///
         /// Note
         /// ----
         /// The action will only be animated if it is animated, **and** the list update itself is
         /// animated. Otherwise, no animation occurs.
-        public var animated : Bool
+        public var animation : ScrollAnimation
         
         /// An additional check you may provide to approve or reject the scroll action.
         public var shouldPerform : (ListScrollPositionInfo) -> Bool
+        
+        /// Called when the list performs the insertion.
+        public var didPerform : (ListScrollPositionInfo) -> ()
     }
 }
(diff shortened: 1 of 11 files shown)


ListableUI/Sources/ListView/ListView.swift
```diff
@@ -869,11 +869,13 @@ public final class ListView : UIView, KeyboardObserverDelegate
             let wasInserted = addedItems.contains(info.insertedIdentifier)
             
             if wasInserted && info.shouldPerform(self.scrollPositionInfo) {
+                
                 /// Only animate the scroll if both the update **and** the scroll action are animated.
                 let bothAnimate = info.animated && animated
                 
                 if let destination = info.destination.destination(with: self.content) {
                     self.scrollTo(item: destination, position: info.position, animated: bothAnimate)
+                    info.didPerform(self.scrollPositionInfo)
```
kylebshr, line 878: If it's animated, should the callback happen after the animation?

kylebshr (approved):

ListableUI/Sources/ListView/ListView.swift
kyleve: Hmm, true.. though not every scroll method has callbacks for this, including the ones we use, I think. Let me play with it.

kylebshr: Yeah not sure it really matters, just a thought - maybe leave it unless it becomes an issue ¯\\\_(ツ)_/¯

kyleve: Looks like wrapping up the animation changes in a UIView.animate gives me the right callback.

ListableUI/Sources/ScrollAnimation.swift
```diff
@@ -0,0 +1,98 @@
...
+    public enum AnimationOptions : Hashable {
+        case curveEaseInOut
+        case curveEaseIn
+        case curveEaseOut
+        case curveLinear
+    }
+}
+
+
+extension Set where Element == ScrollAnimation.AnimationOptions {
```
kyleve, line 70: Note: I tend to do this instead of `OptionSet`, since `OptionSet` requires exposing `rawValue`, which is very... C like

ListableUI/Sources/AutoScrollAction.swift
```diff
@@ -94,17 +96,20 @@ public extension AutoScrollAction
         public var insertedIdentifier : AnyIdentifier
         
         /// The desired scroll position,
-        public var position : ItemScrollPosition
+        public var position : ScrollPosition
         
         /// Should the change be animated.
```
kylebshr, line 101: Might want to update the doc comment (and is the Note still accurate?)

ListableUI/Sources/ScrollAnimation.swift
```diff
@@ -0,0 +1,98 @@
...
+    
+    /// No animation is performed.
+    case none
+    
+    /// A default animation is performed. This is the same as `.custom()`.
+    case `default`
+    
+    /// A custom animation is performed.
+    /// The default parameters are 0.25 seconds and `.curveEaseInOut` animation curve.
+    case custom(duration : TimeInterval = 0.25, options : Set<AnimationOptions> = .default)
```
kylebshr, line 22: IMO `custom` should have a closure that takes in a closure (`case custom(performAnimations: (() -> Void) -> Void)`) so you can choose to use `UIViewPropertyAnimator` or `UIView.animate` (I know I always reach for a property animator now, since it has much nicer spring settings).

But then I guess you need to call the completion too. 

Should probably just offer cases for the options in property animator if anyone actually wants those ever.

ListableUI/Sources/ScrollAnimation.swift
```diff
@@ -0,0 +1,98 @@
...
+    /// Performs the provided animations for the `ScrollAnimation`.
+    public func perform(animations : @escaping () -> (), completion : @escaping (Bool) -> () = { _ in })
+    {
+        switch self {
+        case .none:
+            animations()
+            completion(true)
+            
+        case .default:
+            Self.custom().perform(
```
kylebshr, line 42: We could use the property animator extension I made to match the animation iOS presumably uses https://github.com/kyleve/Listable/blob/main/ListableUI/Sources/Internal/UIViewProperyAnimtor+System.swift

kylebshr: (also lol @ the typo in that file name)

kylebshr (approved):

ListableUI/Sources/ScrollAnimation.swift
kyleve: Yeah, this was my thinking here – we could definitely support this, but (a); no one needs it right now, and (b); I don't know how outside of the normal UIView animation blocks, setting the contentOffset responds.

ListableUI/Sources/ScrollAnimation.swift
kyleve: Just a startup name

ListableUI/Sources/AutoScrollAction.swift
kyleve: Ah shit, right.. done

ListableUI/Sources/ScrollAnimation.swift
kyleve: (Fixing file name)

ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -231,24 +235,60 @@ extension ListView
         {
             self.view.updatePresentationState(for: .didEndDecelerating)
         }
+        
+        private var scrollToTopRevertOffset : CGPoint? = nil
```
kyleve, line 239: Rather than this, find the first visible item, and its offset from the top of the screen, and store that? Eg more resilient to content changing.

ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -231,24 +235,60 @@ extension ListView
...
-            case .enabled: return true
+            case .disabled: break
+                
+            case .enabled(let reverts):
+                if let revert = self.scrollToTopRevertOffset {
+                    scrollView.isUserInteractionEnabled = false
+                    
+                    self.scrollToTopRevertOffset = nil
+                     
+                    self.view.revertScrollToTopForStatusBarTap(with: revert) { finished in
```
kyleve, line 253: Verify this matches the other methods constants

ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -242,7 +242,7 @@ extension ListView
         
         func scrollViewDidScrollToTop(_ scrollView: UIScrollView)
         {
-            self.view.updatePresentationState(for: .scrolledToTop)
```
kyleve: Bring this back

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -68,6 +68,7 @@ public final class ListView : UIView, KeyboardObserverDelegate
         self.collectionView.view = self
 
         self.dataSource.presentationState = self.storage.presentationState
+        self.dataSource.environment = self.environment
```
kyleve, line 71: Merge conflict?

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -464,7 +497,52 @@ public final class ListView : UIView, KeyboardObserverDelegate
...
+            animation.perform(
+                animations: {
+                    self.collectionView.setContentOffset(contentOffset, animated: false)
+                },
+                completion: completion
+            )
+        }
+    }
+    
+    static let scrollToTopAnimation = ScrollAnimation.spring(duration: 0.4)
```
kyleve: Remove

ListableUI/Sources/ListView/ListView.swift
kyleve: Nope, I think this was just missing

PR #231, opened 2020-11-18
Change the default BG color for dark mode.

kyleve:


ListableUI/Sources/Appearance.swift
```diff
@@ -32,4 +32,22 @@ public struct Appearance : Equatable
         
         configure(&self)
     }
+    
+    /// The default background color for the `Appearance`.
+    public static var defaultBackgroundColor : UIColor {
+        if #available(iOS 13.0, *) {
+            return UIColor { traits in
```
bencochran, line 39: Should this be `UIColor.systemBackgroundColor`?

kyleve: Yeah tbh I'm not sure... It makes me uncomfy that this value could change out from under us. Would rather keep it static; which seems to be what we've learned over the years.

kyleve: I think imma stick with this

PR #245, opened 2020-12-18
Switch to compile time validation for types

kyleve:


ListableUI/Sources/Internal/Notification.Name+SwipeActions.swift
```diff
@@ -1,12 +0,0 @@
-//
```
kyleve, line 1: Dead file, forgot to remove

PR #258, opened 2021-01-22
Rename .list to .table

kyleve:
In preparation for making the `ListLayout` and associate types public, I am go through and doing some long-standing cleanup to various APIs; one of which is a rename of the `.list` layout type – it is now `.table`, which is a bit clearer (since it mimics a table view), and it reduces confusion with the basic protocol for all layouts, `ListLayout`.

After this, I need to go through and give `grid` its own `Layout` type, because it is still currently using layout from the the `table`.

Note: The `horizontal` tests are gone from the table layout, because they no longer exist. I just forgot to clean up the snapshots previously.


ListableUI/Sources/Deprecations.swift
```diff
@@ -40,79 +40,14 @@ import Foundation
 ///
```
kyleve, line 40: Cleaned up this since a ton of the deprecations were hella old

ListableUI/Sources/Layout/Table/TableListLayout.swift
```diff
@@ -133,107 +133,100 @@ public struct ListAppearance : ListLayoutAppearance
 }
 
 
-extension ListAppearance
+extension TableAppearance
 {
-    public enum Item {
```
kyleve, line 138: Just removed a needless layer of nesting on these.

kylebshr:
I didn't realize what was being renamed and panicked for a moment. That being said, isn't `List` much more common for this kind of layout outside of `UITableView`? e.g., SwiftUI's List. I'm fine with this, but maybe the rename should `ListLayout` → `Layout` or something else?

kyleve:
Yeah there's no great answer here – given the `ListView` or `List` is really a collection view style view; the naming gets pushed down a layer. And I think I want to keep `ListLayout` as `ListLayout` to line up with` ListView`, `List`, etc.

kyleve:
(Put I did always find it confusing that a `ListView` or `List` has a `List` layout (what else would it have?), but then also `grid` or `paged` layouts – overload of terminology IMO...)

kylebshr:
mmm that's a really good point, I did find that weird too - I like this 👍

kylebshr (approved):

PR #263, opened 2021-01-26
Fix memory leak which caused populated list views to leak, also prep 0.15.1

kyleve:


ListableUI/Tests/ListView/ListViewTests.swift
```diff
@@ -13,6 +13,38 @@ import XCTest
...
+            listView?.configure { list in
+
+                list("content") { section in
+                    section += TestContent(title: "1")
+                    section += TestContent(title: "2")
+                    section += TestContent(title: "3")
+                }
+            }
+
+            self.waitForOneRunloop()
```
kyleve, line 36: This forces the collection view to layout. For some reason`layoutIfNeeded` doesn't do it.

ListableUI/Sources/Internal/Presentation State/PresentationState.ItemState.swift
```diff
@@ -66,8 +66,8 @@ protocol ItemContentCoordinatorDelegate : AnyObject
 
 public struct ItemStateDependencies
```
kyleve, line 67: This is the fix.

bencochran (approved):

watt (approved):

PR #270, opened 2021-03-01
[SwipeActions] Limit overscrolling to the right

nononoah:
### Summary

Though Listable does not support actions beneath the left of the cell, it is currently possible to swipe the contents of a cell all the way to the right. This leads to unexpected and undesirable visual state.

This PR caps overscroll to the right to 20% of the width of the cell. As a result, we maintain a certain bounciness, but don't allow for wilder translations.

### Demo

**Before**
before
https://user-images.githubusercontent.com/1542313/109521585-78221e00-7a7b-11eb-846c-c04962cf200d.mov

**After**
after
https://user-images.githubusercontent.com/1542313/109521619-7fe1c280-7a7b-11eb-8676-d7b00faf1e4c.mov

### DIFF
--- a/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -137,6 +137,7 @@ extension ItemCell {
 
             guard let configuration = swipeConfiguration else { return }
 
+            let velocity = sender.velocity(in: self).x
             let offsetMultiplier = configuration.numberOfActions == 1 ? 0.5 : 0.7
             let performActionOffset = frame.width * CGFloat(offsetMultiplier)
             let currentSwipeOffset = -contentView.frame.origin.x
@@ -152,14 +153,18 @@ extension ItemCell {
             switch sender.state {
             case .began, .changed:
 
-                let swipeState = SwipeActionState.swiping(willPerformAction: willPerformAction)
-                set(state: swipeState)
+                if swipeState == .closed && velocity > 0 {
+                    // The cell is closed and this is a swipe to the right. Ignore the swipe.
+                    sender.setTranslation(.zero, in: self)
+                } else {
+                    let swipeState = SwipeActionState.swiping(willPerformAction: willPerformAction)
+                    set(state: swipeState)
+                }
 
             case .ended, .cancelled:
 
                 let swipeActionsWidth = configuration.swipeView.swipeActionsWidth
                 let keepOpenOffset = swipeActionsWidth / 2
-                let velocity = sender.velocity(in: self).x
 
                 var swipeState: SwipeActionState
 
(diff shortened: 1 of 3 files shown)


kyleve:
Mind adding a note to the changelog in the root of the repo as well?

nononoah:
> Mind adding a note to the changelog in the root of the repo as well?

Added!

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -68,7 +68,8 @@ extension ItemCell {
             case .swiping:
 
                 let translation = configuration.panGestureRecognizer.translation(in: self)
-                xOriginOffset = contentView.frame.origin.x + translation.x
+                // No actions exist to the left, so limit overscrolling to the right to 20% of the width.
+                xOriginOffset = min(bounds.width / 5.0, contentView.frame.origin.x + translation.x)
```
kylebshr, line 72: Can I suggest an additional change that doesn't allow you _start_ a swipe in this direction, which mimics the built-in swipe actions? (This should stay since you can still swipe this was after starting a swipe). Happy to PR that change myself, or walk you through what I have in mind

kylebshr (approved):

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
nononoah: Thanks for the suggestion,@kylebshr. It definitely improves the feel of the cell. 

I took a shot at implementing this, and arrived at the following:

```swift
case .began, .changed:

    if currentSwipeOffset == 0 && velocity > 0 {
        // The cell is closed and this is a swipe to the right. Ignore the swipe.
        sender.setTranslation(.zero, in: self)
        set(state: .closed)
    } else {
        let swipeState = SwipeActionState.swiping(willPerformAction: willPerformAction)
        set(state: swipeState)
    }
```

`sender.setTranslation(.zero, in: self)` is important so that we don't "jump" when the swipe that begins to the right starts moving back to the left.

`set(state: .closed)` is mostly nonsense (we're already closed), but I wanted to preserve the convention of always setting state in this switch, which perhaps is not necessary.

If you have other ideas or would like the logic spelled differently, please let me know and I'll update before merging.

kylebshr: My one thought is that those could perhaps be implemented in the gesture recognizer itself (`HorizontalGestureRecognizer.swift` I think?), in much the same way that it ignores vertical swipes. That might clean up the logic here, since the gesture will cancel and never start if the velocity is incorrect.

kylebshr (approved):
Awesome, thanks! We'll probably need to update this eventually when we add support for trailing _and/or_ leading actions, but we can tackle that when we get to it!

PR #271, opened 2021-03-01
Fix crash that occurs when transitioning to a zero size view

kyleve:
This fixes an issue that occurs when list view transitions from a sized view, to one with zero area. This happens because when the size transitions to a zero area view, we remove all content – it doesnt mean anything to lay out content that has no possible size. Unfortunately, when updating the visible content, we depend on the collection view's layout, which has not yet updated, and causes a crash due to returning content that no longer exists. The fix here is to return early in this case.

I think there might be a more holistic way to fix this by updating the visible content at a slightly different time, but this does the trick for now. I'll keep playing with it.


kylebshr (approved):

kyleve:
Took a better shot at this here: https://github.com/kyleve/Listable/pull/273

DJBen:
Thank you for the quick fix!

PR #274, opened 2021-03-02
Support blueprint update animations in coordinator updates

kyleve:
This PR updates the somewhat experimental `ItemContentCoordinator` APIs to properly support animations for Blueprint-based cells. It has a few changes:

1. Share the `ViewAnimation` type across both scrolling and item content updates, by renaming `ScrollAnimation` to `ViewAnimation`.
2. Remove the `View`/`view` from the `ItemContentCoordinator` – exposing this makes no sense for blueprint cells, and also, all updates to the item content should be done through apply updates, not through manually poking the view.
3. Update the machinery used to propagate animations from coordinators to actually properly wrap updates in an animation block.
4. Add a demo for showing how to do an "on tap" animation when a row is tapped.

### DIFF
--- a/ListableUI/Sources/Item/ItemContentCoordinator.swift
+++ b/ListableUI/Sources/Item/ItemContentCoordinator.swift
@@ -57,6 +57,9 @@ public protocol ItemContentCoordinator : AnyObject
     /// The type of `ItemContent` associated with this coordinator.
     associatedtype ItemContentType : ItemContent
     
+    /// The item associated with the coordinator.
+    typealias Item = ListableUI.Item<ItemContentType>
+    
     // MARK: Actions & Info
     
     /// The available actions you can perform on the coordinated `Item`. Eg, updating it to a new value.
@@ -68,38 +71,32 @@ public protocol ItemContentCoordinator : AnyObject
     // MARK: Instance Lifecycle
     
     /// Invoked on the coordinator when it is first created and configured.
-    func wasInserted(_ info : Item<ItemContentType>.OnInsert)
+    func wasInserted(_ info : Item.OnInsert)
     
     /// Invoked on the coordinator when its owned item is removed from the list due to
     /// the item, or its entire section, being removed from the list.
     ///
     /// Not invoked during deallocation of a list.
-    func wasRemoved(_ info : Item<ItemContentType>.OnRemove)
+    func wasRemoved(_ info : Item.OnRemove)
     
     /// Invoked on the coordinator when its owned item is moved inside a list due to its
     /// order changing.
     ///
     /// Not invoked when an item is manually re-ordered by a user.
-    func wasMoved(_ info : Item<ItemContentType>.OnMove)
+    func wasMoved(_ info : Item.OnMove)
     
     /// Invoked on the coordinator when an external update is pushed onto the owned `Item`.
     /// This happens when the developer updates the content of the list, and the item is
     /// reported as changed via its `isEquivalent(to:)` method.
-    func wasUpdated(_ info : Item<ItemContentType>.OnUpdate)
+    func wasUpdated(_ info : Item.OnUpdate)
     
     // MARK: Visibility & View Lifecycle
-    
-    /// The view type associated with the item.
-    typealias View = ItemContentType.ContentView
-    
-    /// The view, if any, currently used to display the item.
-    var view : View? { get set }
 
     /// Invoked when the list is about to begin displaying the item with the given view.
-    func willDisplay(with view : View)
+    func willDisplay()
 
     /// Invoked when the list is about to complete displaying the item with the given view.
-    func didEndDisplay(with view : View)
+    func didEndDisplay()
     
     // MARK: Selection & Highlight Lifecycle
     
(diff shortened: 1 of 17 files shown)


gilgtm (approved):

ListableUI/Tests/Internal/Presentation State/PresentationState.ItemStateTests.swift
```diff
@@ -402,30 +402,18 @@ fileprivate struct TestContent : ItemContent, Equatable
...
-        }
-        
-        var willDisplay_calls = [View]()
+        var willDisplay_calls = [Void]()
 
-        func willDisplay(with view : View)
+        func willDisplay()
         {
-            self.willDisplay_calls.append(view)
+            self.willDisplay_calls.append(())
```
kylebshr, line 409: I don't really understand what these arrays are doing, what's the purpose?

ListableUI/Sources/ViewAnimation.swift
```diff
@@ -8,45 +8,44 @@
...
-    /// A custom animation is performed.
+    /// A `UIView.animate(...)` animation is performed.
     /// The default parameters are 0.25 seconds and `.curveEaseInOut` animation curve.
-    case custom(duration : TimeInterval = 0.25, options : Set<AnimationOptions> = .default)
+    case animated(TimeInterval = 0.25, options : Set<AnimationOptions> = .default)
     
-    case spring(duration : TimeInterval = 0.25, timing : UISpringTimingParameters = .init())
+    /// A spring based animation is performed.
+    /// The default parameters are 0.25 seconds and `UISpringTimingParameters()` timing.
+    case spring(TimeInterval = 0.25, timing : UISpringTimingParameters = .init())
```
kylebshr, line 27: Using `UISpringTimingParameters` with a `UIViewPropertyAnimator` ignores the duration and uses the spring dampening to determine it. It's so dumb that the API takes in a duration, but it's completely ignored. I don't think it makes sense to expose a duration here.

ListableUI/Sources/ViewAnimation.swift
```diff
@@ -8,45 +8,44 @@
...
-    /// A default animation is performed. This is the same as `.custom()`.
-    case `default`
+    /// A default animation is performed. This is the same as `.animated()`.
+    public static var `default` : Self = .animated()
     
-    /// A custom animation is performed.
+    /// A `UIView.animate(...)` animation is performed.
     /// The default parameters are 0.25 seconds and `.curveEaseInOut` animation curve.
-    case custom(duration : TimeInterval = 0.25, options : Set<AnimationOptions> = .default)
+    case animated(TimeInterval = 0.25, options : Set<AnimationOptions> = .default)
```
kylebshr, line 23: Is this missing a parameter name? Or TIL about this syntax

ListableUI/Sources/Item/ItemContentCoordinator.swift
```diff
@@ -141,28 +138,41 @@ public extension ItemContentCoordinator
...
+        after delay: TimeInterval = 0,
+        update : (inout Item<Content>) -> ()
+    ) {
         var new = self.currentProvider()
         
         update(&new)
         
-        self.update(animated: animated, new)
+        if delay > 0 {
+            Timer.scheduledTimer(withTimeInterval: delay, repeats: false) { _ in
```
kylebshr, line 170: Does this timer need to be retained? I can never remember. Separately, does it need to be canceled if another update happened? e.g., what if update A happened with a 1s delay, then B with 0.3s, would be occur and then get reverted back to A?

kylebshr (approved):
I think this looks good other than a timer/race condition question

ListableUI/Tests/Internal/Presentation State/PresentationState.ItemStateTests.swift
kyleve: It's for tracking the number of calls to the func during testing, so I can make sure the number of calls is right. It could also just be a counter since the type here is `Void`, but I kept it like this since thats what the rest of the file does.

ListableUI/Sources/ViewAnimation.swift
kyleve: Ahh right, will fix

ListableUI/Sources/ViewAnimation.swift
kyleve: Not needed!

ListableUI/Sources/Item/ItemContentCoordinator.swift
kyleve: It doesn't need to be retained, it gets retained by the Runloop itself, and then fires and then is deallocated.

I didn't figure any queueing / collapsing of info was useful here, since it's a lot of extra behaviour, and might not be expected.

Good point about the disparate delays – I guess I should update the backing value immediately; OR alternatively, not pull the current value until the block executes (that actually seems better?)

kyleve: Going to go with the latter:

```
    public func update(
        animation: ViewAnimation = .default,
        after delay: TimeInterval = 0,
        update : @escaping (inout Item<Content>) -> ()
    ) {
        if delay > 0 {
            Timer.scheduledTimer(withTimeInterval: delay, repeats: false) { _ in
                var new = self.currentProvider()
                update(&new)
                
                self.updateCallback(new, animation)
            }
        } else {
            var new = self.currentProvider()
            update(&new)
            
            self.updateCallback(new, animation)
        }
    }
```

kyleve: Done

ListableUI/Sources/ViewAnimation.swift
kylebshr: Swift is so weird in some ways

PR #277, opened 2021-03-08
[ListActions] Add `scrollToSection`

lokae0:
Add capability to scroll to the section with the provided identifier, with the provided positioning.

The list will first attempt to scroll to the section header. If no header is present, it'll scroll to the first item instead. If neither header nor items are found, the list will fallback to the footer.

Also:
* Renamed the `firstIndexPath` function in `Content.swift` to differentiate between querying with an `itemIdentifier` and a `sectionIdentifier`.
* Cherry-picked @kyleve's [commit](https://github.com/kyleve/Listable/pull/276) to remove `SectionInfo` in favor of using identifier directly on `Section`.

Example usage: 
```
listActions.scrolling.scrollTo(
    section: Identifier<Section>(5),
    position: ScrollPosition(position: .centered)
)
```

Produces the following behavior:

listable_demo

### DIFF
--- a/ListableUI/Sources/Content.swift
+++ b/ListableUI/Sources/Content.swift
@@ -148,11 +148,11 @@ public struct Content
     /// Returns the first `IndexPath` for the contained `Item` with the given `AnyIdentifier`,
     /// if it can be found. If nothing is found, nil is returned.
     /// If you have multiple `Item`s with the same identifier, the first one will be returned.
-    public func firstIndexPathFor(itemIdentifier : AnyIdentifier) -> IndexPath?
+    public func firstIndexPathForItem(with identifier : AnyIdentifier) -> IndexPath?
     {
         for (sectionIndex, section) in self.sections.enumerated() {
             for (itemIndex, item) in section.items.enumerated() {
-                if item.identifier == itemIdentifier {
+                if item.identifier == identifier {
                     return IndexPath(item: itemIndex, section: sectionIndex)
                 }
             }
@@ -161,14 +161,14 @@ public struct Content
         return nil
     }
 
-    /// Returns the first `IndexPath` for the contained `Section` with the given `AnyIdentifier`,
+    /// Returns the first index position for the contained `Section` with the given `AnyIdentifier`,
     /// if it can be found. If nothing is found, nil is returned.
     /// If you have multiple `Section`s with the same identifier, the first one will be returned.
-    public func firstIndexPathFor(sectionIdentifier : AnyIdentifier) -> IndexPath?
+    public func firstIndexForSection(with identifier : AnyIdentifier) -> Int?
     {
         for (sectionIndex, section) in self.sections.enumerated() {
-            if section.identifier == sectionIdentifier {
-                return IndexPath(item: 0, section: sectionIndex)
+            if section.identifier == identifier {
+                return sectionIndex
             }
         }
 
--- a/ListableUI/Sources/Content.swift
+++ b/ListableUI/Sources/Content.swift
@@ -148,16 +148,30 @@ public struct Content
     /// Returns the first `IndexPath` for the contained `Item` with the given `AnyIdentifier`,
     /// if it can be found. If nothing is found, nil is returned.
     /// If you have multiple `Item`s with the same identifier, the first one will be returned.
-    public func firstIndexPath(for identifier : AnyIdentifier) -> IndexPath?
+    public func firstIndexPathFor(itemIdentifier : AnyIdentifier) -> IndexPath?
     {
         for (sectionIndex, section) in self.sections.enumerated() {
             for (itemIndex, item) in section.items.enumerated() {
-                if item.identifier == identifier {
+                if item.identifier == itemIdentifier {
                     return IndexPath(item: itemIndex, section: sectionIndex)
                 }
             }
         }
         
+        return nil
+    }
+
+    /// Returns the first `IndexPath` for the contained `Section` with the given `AnyIdentifier`,
+    /// if it can be found. If nothing is found, nil is returned.
+    /// If you have multiple `Section`s with the same identifier, the first one will be returned.
+    public func firstIndexPathFor(sectionIdentifier : AnyIdentifier) -> IndexPath?
+    {
+        for (sectionIndex, section) in self.sections.enumerated() {
+            if section.identifier == sectionIdentifier {
+                return IndexPath(item: 0, section: sectionIndex)
+            }
+        }
+
         return nil
     }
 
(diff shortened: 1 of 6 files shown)


ListableUI/Sources/ListActions.swift
```diff
@@ -133,6 +133,34 @@ public final class ListActions {
                 completion: completion
             )
         }
+
+        ///
+        /// Scrolls to the section with the provided identifier, with the provided positioning.
+        /// If there is more than one section with the same identifier, the list scrolls to the first.
+        /// If the section has content and is contained in the list, true is returned. If not, false is returned.
+        ///
+        /// The list will first attempt to scroll to the section header. If no header is present, it'll scroll to the
```
kyleve, line 142: Rather than this implicit ordering, thoughts on making it explicit:

```
public func scrollTo(
	position : SectionPosition,
    section : AnyIdentifier,
    scrollPosition : ScrollPosition,
    animation : ScrollAnimation = .none,
    completion : @escaping ScrollCompletion = { _ in }
) -> Bool

enum SectionPoint {
	case top // would start at the header and work down
	case bottom // would start at the footer and work up
}
```

?

Not sure how useful this is, but wanted to raise. Another option would be something like:

```
enum SectionPosition {
   case header
   case content
   case footer
}
```

And just bail if the section doesn't have an associated header/content/footer or something; though I can also see that being more fragile, especially when lists have interspersed sections with and without headers.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -448,6 +449,86 @@ public final class ListView : UIView, KeyboardObserverDelegate
...
+
+        // Make sure the section identifier is valid and the section has content.
+
+        guard let sectionPath = storageContent.firstIndexPathFor(sectionIdentifier: section),
+              layoutContent.sections[sectionPath.section].all.isEmpty == false
+        else {
+            return false
+        }
+
+        let headerFrame = layoutContent.sections[sectionPath.section].header.defaultFrame
```
kyleve, line 480: Might want to make this optional based on if the `header.isPopulated`, I believe? Otherwise IIRC the frame is full width but zero height. better than the `height != 0` check below.

kyleve, line 480: Also before you call this, I believe you need to prep for scroll to with `preparePresentationStateForScroll`. It's possible this layout content isnt yet paged in.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -918,15 +1000,10 @@ public final class ListView : UIView, KeyboardObserverDelegate
 
     private func preparePresentationStateForScroll(to toIndexPath: IndexPath, scroll: @escaping () -> Void) -> Bool {
 
-        // Make sure we have a last loaded index path.
-
-        guard let lastLoadedIndexPath = self.storage.presentationState.lastIndexPath else {
```
kyleve, line 923: Moving this logic into the below `if` changes the return value I think, it will now return true instead of false.

kyleve: (I forget if this was me or not in the cherry-picked commit)

kyleve (changes requested):

ListableUI/Sources/Content.swift
```diff
@@ -161,6 +161,20 @@ public struct Content
         return nil
     }
 
+    /// Returns the first `IndexPath` for the contained `Section` with the given `AnyIdentifier`,
+    /// if it can be found. If nothing is found, nil is returned.
+    /// If you have multiple `Section`s with the same identifier, the first one will be returned.
+    public func firstIndexPathFor(sectionIdentifier : AnyIdentifier) -> IndexPath?
```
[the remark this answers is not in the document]
lokae0: Ah yeah, good call. We only need the section `Int` while we're in Content-land. I was using the func above as an example, but the full path is not applicable here. Will update

ListableUI/Sources/ListView/ListView.swift
lokae0: Ah nice,  I didn't realize `isPopulated` is an option. Will update

ListableUI/Sources/ListView/ListView.swift
lokae0: Oof yeah, now this will never return false.

The challenge I was having here was sections without items. Since `lastIndexPath` currently relies on `nonEmptySections`:

```
    public var lastIndexPath : IndexPath?
    {
        let nonEmptySections : [(index:Int, section:SectionState)] = self.sections.compactMapWithIndex { index, _, state in
            return state.items.isEmpty ? nil : (index, state)
        }
        
        guard let lastSection = nonEmptySections.last else {
            return nil
        }
        
        return IndexPath(item: (lastSection.section.items.count - 1), section: lastSection.index)
    }
```

Any suggestions on how to approach this?

ListableUI/Sources/ListActions.swift
kyleve: That sounds good to me!

ListableUI/Sources/ListView/ListView.swift
kyleve: I think you'll want to use a different heuristic here in that case – `lastIndexPath` correctly returns nil if there's no content, since as mentioned above, section headers/footers having "index paths" is kind of a `UICollectionView` artifact.

One thing you could do here is update this method to something like:

```
func preparePresentationStateForScroll(to position: PrepareForScrollPosition, scroll: @escaping () -> Void) -> Bool

enum PrepareForScrollPosition {
   case indexPath(IndexPath)
   case section(Int)
}
```

And then switch on that value for determining how to do the slicing.

lokae0: Awesome, thanks. I'll give this a go

lokae0:
@kyleve PR updated! 

I ended up adding `SectionPosition` as you suggested, rather than two separate functions.

Also added `preparePresentationStateForScrollToSection` which checks if the requested section has been loaded. If not, it'll update the presentation state with the index path for that section.

Finally, added some logic to prevent a section footer from appearing underneath a sticky header when scrolled to. I actually have another PR queued up that will add this behavior to `scrollToItem` as well.

I tested this with large sample sets to ensure the loading behavior was working, as well as various combinations of headers, footers, and items.

Please see latest commit if you just want the diff from what you've already reviewed. Lmk if any further changes are necessary!

kyleve (approved):

kyleve:
thanks! looks good!

kyleve:
(lmk if you want me to merge or if you're planning on pushing anything else)

lokae0:
@kyleve thanks! It's ready to merge

PR #281, opened 2021-03-16
Add automatic RefreshControl offset adjustment behavior

alexisakers:
### Goal
When `RefreshControl.isRefreshing` becomes `true` because of a change of state that is not caused by the user pulling to refresh (for example changing filters or tapping retry from an error cell), we want to be able to display the refresh control to indicate that loading is in progress.

### Implementation Details
- We add an offsetAdjustmentBehavior property to `RefreshControl ` to allow consumers to opt into this behavior 
- Create a `adjustContentOffsetForRefreshControl` method in `PresentationState` to adjust the offset based on the current state of the refresh control
- Call that method from `updatePresentationState` in `ListView` after content updates and before performing any auto scroll actions

### Demo
**iOS 12**

https://user-images.githubusercontent.com/16192914/111357445-b3b60e00-865f-11eb-8b1c-9836820ad898.mov

**iOS 13**

https://user-images.githubusercontent.com/16192914/111358304-90d82980-8660-11eb-932a-d4d7a024ffd6.mov

**iOS 14**

https://user-images.githubusercontent.com/16192914/111358454-b9f8ba00-8660-11eb-84a4-fb2da50c9335.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 ### Fixed
 
 ### Added
+- Add support for adjusting the content offset when the refresh control becomes visible with the `offsetAdjustmentBehavior` property.
+
+Example usage:
+
+```
+list.content.refreshControl = RefreshControl(
+    isRefreshing: isRefreshing,
+    offsetAdjustmentBehavior: .displayWhenRefreshing(animate: true),
+    onRefresh: onRefresh
+)
+```
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Example usage:
 ```
 list.content.refreshControl = RefreshControl(
     isRefreshing: isRefreshing,
-    offsetAdjustmentBehavior: .displayWhenRefreshing(animate: true),
+    offsetAdjustmentBehavior: .displayWhenRefreshing(animate: true, scrollToTop: true),
     onRefresh: onRefresh
 )
 ```
--- a/ListableUI/Sources/Internal/Presentation State/PresentationState.swift
+++ b/ListableUI/Sources/Internal/Presentation State/PresentationState.swift
@@ -276,9 +276,30 @@ final class PresentationState
             let newControl = RefreshControlState(new)
             view.refreshControl = newControl.view
             self.refreshControl = newControl
+            newControl.update(with: new)
         } else if self.refreshControl != nil, new == nil {
             view.refreshControl = nil
             self.refreshControl = nil
+        }
+    }
+
+    internal func adjustContentOffsetForRefreshControl(in view : UIScrollView)
+    {
+        guard
+            let control = refreshControl,
+            control.model.isRefreshing == true
+        else {
+            return
+        }
+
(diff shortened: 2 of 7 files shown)


ListableUI/Sources/Internal/Presentation State/PresentationState.swift
```diff
@@ -276,11 +276,29 @@ final class PresentationState
...
+    internal func adjustContentOffsetForRefreshControl(in view : UIScrollView)
+    {
+        guard let control = refreshControl, control.model.isRefreshing else {
+            return
+        }
+
+        switch control.model.offsetAdjustmentBehavior {
+        case .displayWhenRefreshing(let animate):
+            let contentOffset = CGPoint(x: 0, y: -control.view.frame.height - view.safeAreaInsets.top)
+            view.setContentOffset(contentOffset, animated: animate)
```
kyleve, line 295: As modeled right now, it'll always scroll to the top of the list when refreshing begins, right? Is that desired?

kyleve: (Eg, seems like maybe it should only become visible if the list is already at the top?)

kylebshr: This is more of a product/design question I feel like, but this is generally why I like to avoid using the refresh control to indicate a programmatic refresh. Perhaps this should also be configurable?

alexisakers: For the feature that prompted this change, we do want to scroll to the top when we start refreshing through a workflow action, but it's a good point, I also think it should be configurable.

Would you rather add a `scrollToTop` associated value to `displayWhenRefreshing`, or have a new `scrollToTopWhenRefreshing` case?

kylebshr: I don't have a huge preference, an associated `scrollToTop` makes sense to me!

kylebshr: Would be great to update the demo for this case too!

alexisakers: @kylebshr added the parameter and updated

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
...
 
 ### Added
+- Add support for adjusting the content offset when the refresh control becomes visible with the `offsetAdjustmentBehavior` property.
+
+Example usage:
+
+```
+list.content.refreshControl = RefreshControl(
+    isRefreshing: isRefreshing,
+    offsetAdjustmentBehavior: .displayWhenRefreshing(animate: true),
```
kylebshr, line 13:
```suggestion
    offsetAdjustmentBehavior: .displayWhenRefreshing(animate: true, scrollToTop: true),
```

kylebshr (approved):

CHANGELOG.md
kylebshr: (Small update to match the new API!)

alexisakers: Good catch, thank you

PR #283, opened 2021-03-26
Improve performance of contentSize by not using a list to back it

kyleve:
This PR refactors how `ListView.contentSize` is implemented, by switching to manually allocating a layout and measuring it, instead of using a static measurement / template view. This has significant performance improvement benefits on its own of about 50%, because we're not going through the entire diff and layout loop to update the measurement view – just creating a layout and measuring it.

To allow this, a few changes happened:

1) Updated the list layout APIs to work without a needed `UICollectionView` – you really just need the layout objects and the information about the layout frame, so I cleaned that up. This involved removing references to `UICollectionView`, making the delegate optional.
2) Move `ListSizing` to the `BlueprintUILists` module – it's only used there anyway.
3) Some small cleanup things as I went.


BlueprintUILists/Sources/List.swift
```diff
@@ -114,13 +114,27 @@ extension List {
...
+            switch self.sizing {
+            case .fillParent:
+                return ElementContent { constraint -> CGSize in
+                    constraint.maximum
+                }
+                
+            case .measureContent(let key, let limit):
+                return ElementContent(
+                    measurementCachingKey: {
+                        if let key = key {
```
livings124, line 126: My instinct is to make this a guard.

kyleve: I tend to do guard if the early return isnt mirrored to the rest of the method – eg an early bail. Since these are the same for if let = nil vs not, I do if/else.

ListableUI/Tests/Internal/Presentation State/PresentationState.ItemStateTests.swift
```diff
@@ -402,14 +402,14 @@ fileprivate struct TestContent : ItemContent, Equatable
             self.wasRemoved_calls.append(())
```
kyleve, line 402: Fixes build warnings in tests

BlueprintUILists/Sources/ListSizing.swift
```diff
@@ -0,0 +1,80 @@
...
+/// ││         ││
+/// ││         ││
+/// │└─────────┘│
+/// └───────────┘
+///
+/// .measureContent
+/// ┌───────────┐
+/// │           │
+/// │           │
+/// │┌─────────┐│
```
kylebshr, line 38: This is pretty silly, feel free to leave as-is, but do you think top-aligning illustrates what will actually happen better?

kylebshr (approved):

BlueprintUILists/Sources/ListSizing.swift
kyleve: I'm going to ignore this given it was the same as before in the old file and I want to get this in SPOS tonight

kylebshr:
Please do 😂

PR #286, opened 2021-04-02
Make `TableListLayout` respect padding by default

zradke:
Previously while you could set padding on the `TableListLayout`, that
padding would be equally distributed to the left and right rather than
respecting any difference between the two. Even if a custom width was
set on the section or items, if the center alignment was used the
resulting attributes would be globally centered versus centered in the
available width between the padding.

This change resolves the issue on two fronts. First, the
`CustomWidth.Alignment.center` now center aligns between the provided
padding. Second, because the `CustomWidth.default` will center align
completely ignoring padding, this change introduces a root `CustomWidth`
during the layout which all other components merge against to calculate
their origin. This allows the correct center alignment with respect to
padding to cascade down throughout the attributes until a component
overrides the width with something else.


kyleve:
Looking at the updated `200.0 x 700.0.snapshot.png` – I think something is a bit wonky. The only spacing provided on that table is the table's entire padding of `UIEdgeInsets(top: 10.0, left: 20.0, bottom: 30.0, right: 40.0)`, but it looks like the rows and section footers are getting inset an extra bit?

ListableUI/Sources/Layout/Table/TableListLayout.swift
```diff
@@ -459,11 +459,13 @@ final class TableListLayout : ListLayout
...
         let viewWidth = context.viewBounds.width
         
-        let rootWidth = TableAppearance.Layout.width(
-            with: viewSize.width,
+        let rootWidth = CustomWidth.custom(CustomWidth.Custom(
             padding: HorizontalPadding(left: layout.padding.left, right: layout.padding.right),
-            constraint: layout.width
-        )
+            width: layout.width,
+            alignment: .center
```
kyleve, line 465: I assume / think the actual value passed here doesn't matter, since we're measuring within a container of the same width, right? Eg, no other place we should be pulling this value from?

kyleve (approved):

kyleve:
Looks great, thank you for fixing! Just one quick last question.

ListableUI/Sources/Layout/Table/TableListLayout.swift
zradke: Not that I can think of. I didn't see an alignment option in the layout itself. I chose `.center` because it most closely matched the behavior of `CustomWidth.default`, with the addition of considering the padding.

kyleve: 👍

kyleve:
@zradke I added you as a collaborator for this repo, so once you accept you should be able to merge (sorry, didnt realize you weren't one already).

PR #289, opened 2021-04-19
Fix additional top padding between table header and first section

zradke:
This change fixes the addition of the top padding between the header and
the first section of the `TableListLayout`, instead letting only the
`headerToFirstSectionSpacing` determine the spacing between the header
and the first section.


ListableUI/Sources/Layout/Table/TableListLayout.swift
```diff
@@ -499,8 +499,8 @@ final class TableListLayout : ListLayout
         
         performLayout(for: self.content.header) { header in
             let hasListHeader = self.content.header.isPopulated
-            let headerWith = header.layouts.table.width.merge(with: rootWidth)
```
kyleve, line 502: oops

kyleve (approved):
Thanks!

PR #292, opened 2021-05-26
Support reordering between sections, support reordering validation

kyleve:
### Added

- [Reordering between multiple sections is now supported](https://github.com/kyleve/Listable/pull/292).
- [Introduced type safe access to `Section` content following reorder events](https://github.com/kyleve/Listable/pull/292). See `Section.filtered`.
- [`ListStateObserver.onItemReordered` was added](https://github.com/kyleve/Listable/pull/292) to observe reorder events at a list-wide level.
- [`ListLayout` was extended](https://github.com/kyleve/Listable/pull/292) to allow customization of in-progress moves. Note that `ListLayout` is not yet public.

### Changed

- [`Reordering` has been renamed to `ItemReordering`, and a new `SectionReordering` has been introduced](https://github.com/kyleve/Listable/pull/292). This allows finer-grained control over validating reorder events at both the item level and section level.
- [`ListReorderGesture` and `ItemReordering.GestureRecognizer` have been heavily refactored](https://github.com/kyleve/Listable/pull/292) to reduce visibility of internal state concerns.
- [`Item.identifier` has been renamed to `Item.anyIdentifier`](https://github.com/kyleve/Listable/pull/292). The new `Item.identifier` property is now a fully type safe identifier.
- [`ReorderingActions` was refactored](https://github.com/kyleve/Listable/pull/292) to expose less public state and ease use in UIView-backed list elements.
- [`Identifier<Represented>` is now `Identifier<Represented, Value>`; eg `Identifier<MyContent, UUID>`](https://github.com/kyleve/Listable/pull/292). This is done to support reacting to reordering events in a more type safe manner, and to make `Identifier` creation more type safe. This is a large breaking change.
- [Changed how `identifier`s for `ItemContent` are represented](https://github.com/kyleve/Listable/pull/292). `ItemContent` now returns a an identifier of a specific `IdentifierType` (eg, `String`, `UUID`, etc), which is then assembled into an `Identifier` by the containing item. Additional APIs have been added for creating `Identifier`s in a more type safe manner. This is a large breaking change.


ListableUI/Sources/Layout/CollectionViewLayout.swift
```diff
@@ -190,22 +192,49 @@ final class CollectionViewLayout : UICollectionViewLayout
         self.neededLayoutType.merge(with: context)
     }
     
+    override func invalidationContext(
```
kyleve, line 195: Might not need this at all?

ListableUI/Sources/Layout/CollectionViewLayout.swift
```diff
@@ -190,22 +192,49 @@ final class CollectionViewLayout : UICollectionViewLayout
...
+        forInteractivelyMovingItems targetIndexPaths: [IndexPath],
+        withTargetPosition targetPosition: CGPoint,
+        previousIndexPaths: [IndexPath],
+        previousPosition: CGPoint
+    ) -> UICollectionViewLayoutInvalidationContext
+    {
+        self.isReordering = true
```
kyleve, line 202: Centralize this onto `invalidate(with:)` via a property on the `InvalidationContext`

ListableUI/Sources/Layout/CollectionViewLayout.swift
```diff
@@ -190,22 +192,49 @@ final class CollectionViewLayout : UICollectionViewLayout
...
+        return context
+    }
+    
     override func invalidationContextForEndingInteractiveMovementOfItems(
         toFinalIndexPaths indexPaths: [IndexPath],
         previousIndexPaths: [IndexPath],
         movementCancelled: Bool
     ) -> UICollectionViewLayoutInvalidationContext
     {
+        isReordering = false
```
kyleve, line 222: Centralize this onto `invalidate(with:)` via a property on the `InvalidationContext`

ListableUI/Sources/Layout/CollectionViewLayout.swift
```diff
@@ -28,6 +28,8 @@ final class CollectionViewLayout : UICollectionViewLayout
         }
```
kyleve, line 28: In this file: Validate that we're not setting the `neededLayoutType` until the index paths of the item actually change.

kyleve, line 28: I think this needs to loop over the moved index paths; not just the first one:

```
    override func invalidateLayout(with context: UICollectionViewLayoutInvalidationContext)
    {
        let view = self.collectionView!
        let context = context as! InvalidationContext
        
        super.invalidateLayout(with: context)
        
        // Handle Moved Items
        
        if
            let from = context.previousIndexPathsForInteractivelyMovingItems,
            let to = context.targetIndexPathsForInteractivelyMovingItems
        {
            let from = from[0]
            let to = to[0]
```

kyleve, line 28: Can I remove the `liveIndexPath` stuff?

kyleve:
Still crashes:
- Go to reordering demo
- Move one item from last section to first section
- Drag any item from first section down to bottom of last section (don't let it go)
- Move it up towards the first section
- Crash (collection view tries to look up (1,1))

ListableUI/Sources/Identifier.swift
```diff
@@ -77,6 +77,10 @@ public final class Identifier<Represented> : AnyIdentifier
 ///
 public class AnyIdentifier : Hashable, CustomDebugStringConvertible
 {
+    public var base : Any {
```
kyleve, line 80: Need to comment this

ListableUI/Sources/HeaderFooter/HeaderFooterContent.swift
```diff
@@ -128,6 +128,8 @@ public protocol HeaderFooterContent
 
 /// Information about the current state of the content, which is passed to `apply(to:for:with:)`
 /// during configuration and preparation for display.
+///
+/// TODO: Rename to `ApplyHeaderFooterContext`
```
kyleve, line 132: Do this by renaming, then deprecate the old name

ListableUI/Sources/Internal/Presentation State/PresentationState.ItemState.swift
```diff
@@ -461,9 +468,37 @@ extension PresentationState
             }
         }
         
-        func moved(with result : Reordering.Result)
+        func beginReorder(with environment: ListEnvironment) {
```
kyleve, line 471: Document this and the below

ListableUI/Sources/Internal/Presentation State/PresentationState.ItemState.swift
```diff
@@ -461,9 +468,37 @@ extension PresentationState
...
+            }
+            
+            self.isReordering = false
+            
+            UIView.animate(withDuration: 0.15) {
+                self.applyToVisibleCell(with: environment)
+            }
+        }
+        
+        private(set) var isReordering : Bool = false
```
kyleve, line 497: Make sure this only runs if things are NOT cancelled

ListableUI/Sources/Item/ItemContent.swift
```diff
@@ -226,7 +228,11 @@ public struct ApplyItemContentInfo
     public var position : ItemPosition
     
     /// Provides access to actions to handle re-ordering the content within the list.
-    public var reordering : ReorderingActions
+    public var reorderingActions : ReorderingActions
+    
+    /// If the item can be reordered.
+    /// Use this property to determine in your `ItemContent` should render a reorder control.
```
kyleve, line 234: `in your` -> `if your`

ListableUI/Sources/Item/Reordering.swift
```diff
@@ -8,37 +8,225 @@
...
+            case .possible: break
+            case .began:
+                let canStart = self.onStart?()
+                
+                let center = self.view?.firstSuperview(ofType: UICollectionViewCell.self)?.center
+                
+                if let center = center, canStart == true {
+                    self.initialCenter = center
+                } else {
+                    self.state = .cancelled
```
kyleve, line 168: Do I need to do `onEnd`, etc, here?

ListableUI/Sources/Item/Reordering.swift
```diff
@@ -8,37 +8,225 @@
...
+            @unknown default: listableFatal()
+            }
+        }
+    }
+}
+
+
+extension Reordering.Sections {
+    
+    func canMove(from : PresentationState.SectionState, to : PresentationState.SectionState) -> Bool {
```
kyleve, line 190: Test

ListableUI/Sources/Item/Reordering.swift
```diff
@@ -8,37 +8,225 @@
...
+        case .specific(let IDs):
+            return IDs.contains(to.model.identifier)
+        }
+    }
+}
+
+
+extension Reordering {
+    
+    func destination(
```
kyleve, line 208: Test

ListableUI/Sources/Layout/CollectionViewLayout.swift
```diff
@@ -163,21 +165,23 @@ final class CollectionViewLayout : UICollectionViewLayout
...
+                if info.from != info.to {
+                    self.layout.content.move(from: info.from, to: info.to)
+                    self.layout.content.reindexIndexPaths()
+                }
+
+            case .complete(_):
+                break
+
+            case .cancelled(let info):
+                self.layout.content.move(from: info.from, to: info.to)
```
kyleve, line 184: Test this manually

ListableUI/Sources/Layout/ListLayout/ListLayout.swift
```diff
@@ -79,6 +79,17 @@ public protocol AnyListLayout : AnyObject
     )
     
     func setZIndexes()
+    
+    //
+    // MARK: Configuring Reordering
+    //
+    
+    func adjust(
```
kyleve, line 87: Instead of using `UICollectionViewLayoutAttributes` directly here; let's add a struct type that only exposes the props needed.

ListableUI/Sources/Layout/ListLayout/ListLayoutContent.swift
```diff
@@ -198,6 +189,39 @@ public final class ListLayoutContent
         self.sections[to.section].items.insert(info, at: to.item)
     }
     
+    func move(from : [IndexPath], to : [IndexPath])
```
kyleve, line 192: Test

ListableUI/Sources/ListView/ListView.DataSource.swift
```diff
@@ -75,27 +77,34 @@ internal extension ListView
...
         }
         
         func collectionView(
             _ collectionView: UICollectionView,
-            moveItemAt sourceIndexPath: IndexPath,
-            to destinationIndexPath: IndexPath
+            moveItemAt from: IndexPath,
+            to: IndexPath
+        ) {
+            self.storage.moveItem(from: from, to: to)
```
kyleve, line 95: Verify: This is only called when the indexes are actually different, right? Maybe throw in a guard check to be safe against regressions...

ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -180,24 +181,52 @@ extension ListView
...
+            let item = self.layoutManager.stateFor(itemAt: from)
+            
+            guard let reordering = item.anyModel.reordering else {
+                return from
+            }
+            
+            let fromSection = self.presentationState.sections[from.section]
+            let toSection = self.presentationState.sections[to.section]
+            
+            return reordering.destination(
```
kyleve, line 224: Todo: This does not move the item from the from section and pass it to the to section. Should we?

ListableUI/Sources/ReorderingActions.swift
```diff
@@ -18,47 +18,63 @@ public final class ReorderingActions
...
-    public func beginMoving() -> Bool
+    public func start() -> Bool
     {
+        guard let item = self.item else {
+            return false
+        }
+        
         guard self.isMoving == false else {
-            return true
+            return false
```
kyleve, line 28: Maybe should still return true?

ListableUI/Sources/ReorderingActions.swift
```diff
@@ -18,47 +18,63 @@ public final class ReorderingActions
...
-        self.delegate?.updateInteractiveMovementTargetPosition(with: recognizer)
+        guard let item = self.item else {
+            return
+        }
+        
+        self.delegate?.updateReorderTargetPosition(with: recognizer, for: item)
     }
     
-    public func end()
+    public func end(_ cancelled : Bool)
```
kyleve, line 57: Consider making this an enum instead of a bool; less confusing

ListableUI/Sources/ReorderingActions.swift
```diff
@@ -18,47 +18,63 @@ public final class ReorderingActions
...
         }
         
+        guard let item = self.item else {
+            return
+        }
+        
         self.isMoving = false
         
-        self.delegate?.endInteractiveMovement()
+        self.delegate?.endReorder(for: item)
```
kyleve, line 69: Shouldn't this call cancel in the cancel case?

ListableUI/Sources/ReorderingActions.swift
```diff
@@ -18,47 +18,63 @@ public final class ReorderingActions
...
     }
 }
 
 
 protocol ReorderingActionsDelegate : AnyObject
```
kyleve, line 74: Merge end vs cancelled into one method with a param to specify the result?

ListableUI/Sources/Layout/CollectionViewLayout.swift
```diff
@@ -163,21 +165,23 @@ final class CollectionViewLayout : UICollectionViewLayout
...
             
-            if from != to {
-                context.performedInteractiveMove = true
-                self.layout.content.reindexLiveIndexPaths()
+            switch action {
+            case .inProgress(let info):
+                
+                if info.from != info.to {
+                    self.layout.content.move(from: info.from, to: info.to)
+                    self.layout.content.reindexIndexPaths()
```
kyleve, line 177: Move this into the move method

ListableUI/Sources/Layout/CollectionViewLayout.swift
kyleve: How can I test this... automated...ly?

ListableUI/Sources/ListView/ListView.Delegate.swift
kyleve: I don't think it should. Going to leave as is; since the change has not been committed.

ListableUI/Sources/ListView/ListView.LayoutManager.swift
```diff
@@ -22,6 +22,10 @@ extension ListView
             self.collectionView = collectionView
         }
         
+        func stateFor(itemAt indexPath: IndexPath) -> AnyPresentationItemState {
```
kyleve, line 25: `stateForItem(at:`

kyleve: Done

Demo/Sources/Demos/Demo Screens/ReorderingViewController.swift
```diff
@@ -14,95 +14,83 @@ import BlueprintUI
...
 
 
-final class ReorderingViewController : UIViewController
+final class ReorderingViewController : ListViewController
 {
-    let list = ListView()
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        self.navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Reload", style: .plain, target: self, action: #selector(reload))
```
kyleve, line 22: Still need to test this.

kyleve: This doesn't work; but since its an edge case; I'm going to land and work on getting into SPOS. Can follow up w/ this.

ListableUI/Sources/Layout/ListLayout/ListContentLayoutAttributes.swift
```diff
@@ -0,0 +1,93 @@
...
+            
+            self.size = newValue.size
+        }
+    }
+    
+    //
+    // MARK: Transforms & Layout
+    //
+    
+    public var transform: CGAffineTransform
```
kyleve, line 53: Consider:

```
    public enum Transform {
        case affine(CGAffineTransform)
        case transform3d(CATransform3D)
    }
```

ListableUI/Sources/ListView/ListView.CollectionViewChanges.swift
```diff
@@ -25,6 +25,7 @@ internal extension ListView
         var movedItems : [ItemChanges.Moved] = []
                 
         var hasIndexAffectingChanges : Bool {
+            return true
```
kyleve, line 28: - [ ] Revert before merge

ListableUI/Sources/Identifier.swift
```diff
@@ -14,36 +14,59 @@ import Foundation
...
+    /// strongly typed identifiers with the correct `Value` type:
+    /// ```swift
+    /// MyItem.identifier(with: "my-id")  // ✅ OK
+    /// MyItem.identifier(with: 1)        // 🚫 Error: MyItem's IdentifierValue is String.
+    /// ```
+    ///
+    /// You can also read  ``Item.identifier-swift.property`` or ``AnyItem.anyIdentifier-swift.property``
+    /// to get the identifier of an item that has been created in a type safe manner.
+    ///
+    init(_ value : Value)
```
alanjzhang, line 69: I think this needs to be public - here's an [example](https://github.com/squareup/ios-register/blob/aba13e6140ad1e22ad6b538a2b31d35744f304f2/Frameworks/SquareSettings/Impl/Sources/Master/SettingsMasterScreen.swift#L140) of where this is needed. I just pointed my local Podfile to your branch and had to update a few things from the breaking API changes.

alanjzhang, line 69: I think this needs to be public. I just pointed my local Podfile to your branch and had to update a few things from the breaking API changes.

ListableUI/Sources/Layout/Table/TableListLayout.swift
```diff
@@ -398,7 +398,7 @@ final class TableListLayout : ListLayout
     typealias LayoutAppearance = TableAppearance
     
     static var defaults: ListLayoutDefaults {
-        .init(itemInsertAndRemoveAnimations: .top)
+        .init(itemInsertAndRemoveAnimations: .fade)
```
kyleve, line 401: Changing this default; it looks better.

PR #295, opened 2021-06-11
Result Builder Support

kyleve:
Adds support for building list content and section content with result builders, supporting SwiftUI-style syntax. I have a couple possible syntaxes to move forward with, so please check out [this comment](https://github.com/kyleve/Listable/pull/295/files#r739891646) for the options and a brief discussion.

**Note**: I haven't (entirely) converted the demo app to use result builders in this PR to reduce PR size. Will follow up with that since it's basically a mechanical conversion.


Demo/Sources/Demos/BestPractices.swift
```diff
@@ -0,0 +1,134 @@
...
+//
+
+import ListableUI
+import BlueprintUILists
+
+
+/// This file contains examples for best practices for how to build lists and sections.
+enum BestPractices {
+    
+    static func make_list_api_1() -> List
```
kyleve, line 16: This example shows the API if we use the `callAsFunction` syntax (eg, `list { ... }`), whereas the second example below shows the result builders as a second parameter. Thoughts on which one is preferable?

I've explicitly _avoided_ added a version of initializers where there's initializer parameters for header, footer, layout, etc, and instead kept the `configure` blocks, for a couple reasons:

1) Xcode will either autocomplete methods two ways: with _all_ the arguments with default values, or with none of them. This means that if you want to override half the default values, you need to either delete, or add back, half the arguments to methods like `Section` or `ListProperties`' inits (and know which have defaults) – it's sort of a pain in the ass.

2) Many times, people have logic around setting headers, footers, or layouts. If we just provide these as parameters, the logic is harder to put inline (eg you'd need an inline closure).

3) I could also provide options like header, footer, as closure-based parameters (eg, `() -> HeaderFooter`), however due to the ambiguity of the _first_ trailing closure being unnamed, and both header and footer being optional, the compiler can't actually tell which I would be returning.

4) Keeping the `configure` closure, but allowing using result builders to add sections or add items to sections, provides a much easier migration path to use result builders.

ListableUI/Sources/ListProperties.swift
```diff
@@ -195,19 +194,59 @@ public struct ListProperties
...
+    /// ```
+    /// With:
+    /// ```
+    /// ListProperties { list in
+    ///     list.header = ...
+    ///     list.footer = ...
+    ///     ...
+    /// }
+    /// ```
+    public subscript<Value>(dynamicMember keyPath: WritableKeyPath<Content, Value>) -> Value {
```
kyleve, line 219: Another quality of life addition; see comment for what it does.

Demo/Sources/Demos/BestPractices.swift
n8chur: I'm not a fan of the `callAsFunction` syntax, I much prefer the result builder as second parameter approach.

I haven't seen `callAsFunction` used in other APIs provided by Apple so the concept may not be intuitive to most consumers. Also using `callAsFunction` for this here means that you're forced into a deeper layer of nesting, and it's likely not as discoverable.

kyleve: > Had you explored making each of these separate result builders as well? 

I can't do that, due to this: 

> I could also provide options like header, footer, as closure-based parameters (eg, () -> HeaderFooter), however due to the ambiguity of the first trailing closure being unnamed, and both header and footer being optional, the compiler can't actually tell which I would be returning.

Eg:

```
Section("Identifier") {
   /// Is this the header or the footer?
} items: {
   ...
}
```

n8chur: ohh because we have the optional configure closure first you mean? hm

n8chur: Wait actually I don't think I do understand why this is a problem. If `content` (`items`?) is always first (like it is with SwiftUI) and is always required, would that solve the problem?

kyleve: It would, but I generally don't like putting the _longest_ parameter (in terms of LOC) at the top of the initializer, since you end up in this kind of case, where there's little "bits" of config waaay farther down the page:

```
Section("id") {
  MyContent(...)
  MyContent(...)
  MyContent(...)
  MyContent(...)


  switch foo {
  case blah:
      self.other.map {
       MyOtherContent(...)
    }
  }

  ...



   ...
} header: {
   // Usually just 1 or 2 LOC
}
```

kyleve: (Though maybe that's fine if that's the direction Apple is going and it fixes the problem)

n8chur: Yeah IMO we should follow Apple's lead here. Curious how others feel.

kyleve: Tbh seems fine, I'll switch it over to that!

kyleve: Done

Demo/Sources/Demos/DemosRootViewController.swift
```diff
@@ -238,7 +238,7 @@ public final class DemosRootViewController : ListViewController
             })
         }
         
-        list("selection-state") { [weak self] section in
```
kyleve, line 241: Not used in this method

ListableUI/Sources/Section/Section.swift
```diff
@@ -116,31 +116,38 @@ public struct Section
...
-        self.items += content().map { $0.toAnyItem() }
+        self.identifier = Identifier(identifier)
+        
+        self.layouts = .init()
+        self.reordering = .init()
+        
+        self.items = items().map { $0.toAnyItem() }
+        
+        self.header = header()
+        self.footer = footer()
```
n8chur, line 156: Should this just delegate to the other initializer?

Demo/Sources/Demos/BestPractices.swift
n8chur: Nice!

Are we also removing the `callAsFunction` syntax or is that still TBD?

ListableUI/Sources/ListableBuilder.swift
```diff
@@ -0,0 +1,88 @@
...
+    public typealias FinalResult = [ContentType]
+    
+    /// If an empty closure is provided, returns an empty array.
+    public static func buildBlock() -> Component {
+        []
+    }
+    
+    /// Required by every result builder to build combined results from statement blocks.
+    public static func buildBlock(_ components: Component...) -> Component {
+        components.reduce(into: []) { $0 += $1 }
```
kylebshr, line 46: Can just do this - or is `into` more performant?
```suggestion
        components.reduce([], +)
```

ListableUI/Sources/ListableBuilder.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /// With buildEither(first:), enables support for 'if-else' and 'switch' statements by folding conditional results into a single result.
+    public static func buildEither(second component: Component) -> Component {
+        component
+    }
+
+    /// Enables support for 'for..in' loops by combining the results of all iterations into a single result.
+    public static func buildArray(_ components: [Component]) -> Component {
+        components.reduce(into: []) { $0 += $1 }
```
kylebshr, line 76: Same re: simpler reduce

ListableUI/Sources/Section/Section.swift
kyleve: I do this so if I add properties, I'll get a compiler error – delegating to initializers with default values makes it easier to forgot to pass through new parameters.

Demo/Sources/Demos/BestPractices.swift
kyleve: I've removed the new ones – we had / have existing ones for doing `list("foo") { section in ... }` for sections

ListableUI/Sources/ListableBuilder.swift
```diff
@@ -0,0 +1,88 @@
...
+/// ## Links & Videos
+/// https://github.com/apple/swift-evolution/blob/main/proposals/0289-result-builders.md
+/// https://developer.apple.com/videos/play/wwdc2021/10253/
+/// https://www.swiftbysundell.com/articles/deep-dive-into-swift-function-builders/
+/// https://www.avanderlee.com/swift/result-builders/
+///
+/// ### Note
+/// Most comments on methods come from the result builders SE proposal.
+///
+@resultBuilder public enum ListableBuilder<ContentType> {
```
kylebshr, line 28: Pretty nitpicky but how about `ListBuilder`?

kylebshr (approved):
Pretty neat how small this diff is. I wonder why Apple didn't throw a generic `Builder` into Swift

ListableUI/Sources/ListableBuilder.swift
kyleve: This is slightly faster – so I usually use this instead

ListableUI/Sources/ListableBuilder.swift
kyleve: It's used for more than `List` – sections and also swipe actions, so I just wanted to namespace it compared to the Blueprint, etc, one.

kylebshr: Sounds good, doesn't matter much!

PR #306, opened 2021-07-01
[Blocked] Move to blueprint which includes environment on measure

kyleve:


BlueprintUILists/Sources/List.swift
```diff
@@ -76,79 +76,42 @@ public struct List : Element
...
+                    if let key = key {
+                        return MeasurementCachingKey(type: Self.self, input: key)
+                    } else {
+                        return nil
+                    }
+                }()
+            ) { context -> CGSize in
+                ListView.contentSize(
+                    in: context.constraint.maximum,
+                    for: self.properties,
```
kyleve, line 97: Need to pass the env through here

PR #309, opened 2021-07-26
Mosaic layout added.

gabreho:
#### Goal
We need to customize the size and placement of each tile in a grid layout, and we need two flavours: pages (number of rows) and infinite scroll.

#### Implementation Details
`RetailGridListLayout` and `RetailGridAppearance` added. ItemLayout allows us to define specific coordinates and sizes.

#### Demo
iOS 12
grid


Demo/Sources/Demos/Demo Screens/MosaicViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+//  Copyright © 2021 Kyle Van Essen. All rights reserved.
+//
+
+import UIKit
+
+import ListableUI
+import BlueprintUILists
+
+
+final class MosaicViewController : UIViewController
```
kyleve, line 15: Minor nit / FYI: You can depend on `ListViewController` and then just override `public override func configure(list: inout ListProperties)`

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
```diff
@@ -0,0 +1,302 @@
...
+
+
+extension MosaicAppearance {
+    
+    public struct ItemLayout : ItemLayoutsValue {
+        
+        public let origin: Origin
+        public let size: Size
+        
+        public struct Origin {
```
kyleve, line 97: Nit: Might be nice to make this Equatable?

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
```diff
@@ -0,0 +1,302 @@
...
+ 
+    let appearance : Appearance
+    let behavior : Behavior
+    
+    let content : ListLayoutContent
+    
+    var scrollViewProperties: ListLayoutScrollViewProperties {
+        .init(
+            isPagingEnabled: self.layoutAppearance.layout.rows != .infinite,
+            contentInsetAdjustmentBehavior: .never,
```
kyleve, line 234: To confirm: This will never adjust for safe area insets; intentional?

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
```diff
@@ -0,0 +1,302 @@
...
+                 
+                item.x = frame.origin.x
+                item.y = frame.origin.y
+                item.size = frame.size
+                
+                lastContentMaxY = max(lastContentMaxY, item.y + item.size.height + layout.padding.bottom)
+            }
+        }
+        
+        if scrollViewProperties.isPagingEnabled {
```
kyleve, line 295: This feels like the wrong use of `isPagingEnabled`; it controls where the scroll position settles; but otherwise shouldn't affect layout.

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
```diff
@@ -0,0 +1,302 @@
...
+
+                oneByOneTile = CGSize(width: availableWidth / CGFloat(columns), height: availableHeight / CGFloat(rowCount))
+            case .infinite:
+                oneByOneTile = CGSize(width: availableWidth / CGFloat(columns), height: availableWidth / CGFloat(columns))
+                tileHeight = availableWidth / CGFloat(columns) * size.height.value
+            }
+            
+            tileHeight += (size.height.value - 1) * itemSpacing
+            
+            let interPageSpacing = CGFloat(pageIndex) * pagePadding.top + pagePadding.top
```
kyleve, line 197: Is `pagePadding.top + pagePadding.top` right here; or should one of those be bottom?

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
```diff
@@ -0,0 +1,302 @@
...
+//  ListableUI
+//
+//  Created by Gabriel Hernandez Ontiveros on 2021-07-23.
+//
+
+import Foundation
+
+extension LayoutDescription
+{
+    public static func mosaic(_ configure : @escaping (inout MosaicAppearance) -> () = { _ in }) -> Self
```
kyleve, line 12: I don't love the generic name; maybe let's go with exactly what its used for for now? Thoughts?

kyleve:
Looks really good; nice work! Not a lot of feedback tbh.

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
gabreho: will do!

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
gabreho: I am not sure I follow. This is where I change between paged and infinite scrolling by updating `lastContentMaxY`.

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
gabreho: yeah, I am open to changing the name, but I couldn't come up with a good one, have any ideas? this is for the visual browse (new catalog browser) experience for retail, but RST also is planning to use it.

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
gabreho: not really, I will change it.

Demo/Sources/Demos/Demo Screens/MosaicViewController.swift
gabreho: I used UIViewController to have access to the list view and change the scrolling behaviour, similar to `CustomLayoutsViewController` and how it changes the layout type.

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
gabreho: ah, update: this is correct. `CGFloat(pageIndex) * pagePadding.top` calculates the number of top paddings I need to account for previous pages (for first page would be 0) and then I need to add the current top padding.

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
gabreho: `RetailGridListLayout` and `retailGrid`?

Demo/Sources/Demos/Demo Screens/MosaicViewController.swift
gabreho: Update: I updated to a `ListViewController`, I wasn't thinking in declarative language. but now the issue is that the list doesn't respect the safe area for the navigation bar, is there a way to configure that?

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
kyleve: that seems good to me!

ListableUI/Sources/Layout/Mosaic/MosaicListLayout.swift
kyleve: I guess I mean: I would check `self.layoutAppearance.layout.rows != .infinite`, rather than `isPagingEnabled`, since it makes the intent clearer

Demo/Sources/Demos/Demo Screens/MosaicViewController.swift
kyleve: Is your layout implementation reading the safe area when positioning the first item?

kyleve: (I forget if that's required, but that could be it)

kyleve: Actually nvm, I don't think that is required. This "just works" for the table/list layout; so maybe something different between that impl and your layout?

ListableUI/Sources/ListView/ListView+ContentSize.swift
```diff
@@ -79,7 +79,7 @@ extension ListView
 
         layout.layout(
             delegate: nil,
-            in: .init(viewBounds: CGRect(origin: .zero, size: fittingSize))
+            in: .init(viewBounds: CGRect(origin: .zero, size: fittingSize), safeAreaInsets: .zero)
```
gabreho, line 82: @kyleve I assume `.zero` is safe since this functions calculates a size based in a given value?

Demo/Sources/Demos/Demo Screens/MosaicViewController.swift
gabreho: Updated with ListViewController

ListableUI/Sources/Layout/Retail Grid/RetailGridListLayout.swift
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        public mutating func set(with block : (inout Layout) -> ())
+        {
+            var edited = self
+            block(&edited)
+            self = edited
+        }
+        
+        internal static func width(
```
kyleve, line 79: Used?

ListableUI/Sources/ListView/ListView+ContentSize.swift
kyleve: Yep! This is what I'd expect here

kyleve (approved):
Looks really good! Nicely done.

ListableUI/Sources/Layout/Retail Grid/RetailGridListLayout.swift
gabreho: @kyleve removed

PR #311, opened 2021-08-05
Enable horizontal scroll actions.

gabreho:
#### Goal
Currently scrolling to an item using ListAction doesn't work with horizontal orientation because `ScrollPosition.Position` doesn't cover all available UIKit options.

#### Implementation Details
We return top, bottom and centeredVerically for vertical layouts and left, right and centeredHorizontally for horizontal.

### DIFF
--- a/ListableUI/Sources/ScrollPosition.swift
+++ b/ListableUI/Sources/ScrollPosition.swift
@@ -49,11 +49,20 @@ extension ScrollPosition
         /// Scroll the item to the bottom of the list.
         case bottom
         
-        var UICollectionViewScrollPosition : UICollectionView.ScrollPosition {
-            switch self {
-            case .top: return .top
-            case .centered: return .centeredVertically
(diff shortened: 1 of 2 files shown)


ListableUI/Sources/ScrollPosition.swift
```diff
@@ -51,9 +51,9 @@ extension ScrollPosition
         
         var UICollectionViewScrollPosition : UICollectionView.ScrollPosition {
```
kyleve, line 52: I wonder if it makes sense to have this take in a `LayoutDirection` and then we return either/or; instead of both? Thoughts? Eg:

```
func toUICollectionViewScrollPosition(with direction : LayoutDirection) -> UICollectionView.ScrollPosition
```

gabreho: updated

kyleve (approved):

PR #288, opened 2021-04-10
Support not updating Items and HeaderFooters at all if they are equivalent

kyleve:
This PR updates `ItemContent` and `HeaderFooterContent` to support not updating their on-screen content if they are equivalent. This is done by introducing `ReappliesToVisibleView`, with the option `.always` and `.ifNotEquivalent`.

PR #321, opened 2021-08-31
Fix supplementary reuse bug by ensuring we track the right view instance in our displayed views cache

kyleve:
This PR fixes a bug introduced in #288 (well sort of, 288 uncovered the actual bug):

A long time ago, a method argument was renamed from `view` to `anyView`, which meant we were no longer shadowing the `view` variable on the `Delegate`:

```
        func collectionView(
            _ collectionView: UICollectionView,
                                         // Was renamed from `view`.
            willDisplaySupplementaryView anyView: UICollectionReusableView,
            forElementKind kindString: String,
            at indexPath: IndexPath
            )
        {
            let container = anyView as! SupplementaryContainerView
            let kind = SupplementaryKind(rawValue: kindString)!
            
            let headerFooter : PresentationState.HeaderFooterViewStatePair = ...

            headerFooter.willDisplay(view: container)

          - self.displayedSupplementaryItems[ObjectIdentifier(view)] = headerFooter
          + self.displayedSupplementaryItems[ObjectIdentifier(container)] = headerFooter
```

So the fix here was to properly pass the `ObjectIdentifier` of the `container`, not `view` (`self.view` is actually the `ListView`).

Without this change, `didEndDisplay()` was never invoked on the header/footer (because `displayedSupplementaryItems` is used to track what is being used by the collection view), meaning we never cleared out values during view reuse, which broke the next time a view was reused if the view type was different.

### DIFF
--- a/ListableUI/Sources/Internal/SupplementaryContainerView.swift
+++ b/ListableUI/Sources/Internal/SupplementaryContainerView.swift
@@ -64,6 +64,8 @@ final class SupplementaryContainerView : UICollectionReusableView
         ) as! SupplementaryContainerView
         
         view.reuseCache = reuseCache
+        
+        // TODO: We need to also set this during on-screen updates; I don't think it's updated correctly now.
         view.environment = environment
         
         return view
--- a/ListableUI/Sources/Internal/SupplementaryContainerView.swift
+++ b/ListableUI/Sources/Internal/SupplementaryContainerView.swift
@@ -73,7 +73,7 @@ final class SupplementaryContainerView : UICollectionReusableView
     // MARK: Content
     //
     
-    var headerFooter : AnyPresentationHeaderFooterState? {
+    weak var headerFooter : AnyPresentationHeaderFooterState? {
         didSet {
             guard oldValue !== self.headerFooter else {
                 return
--- a/ListableUI/Sources/Internal/SupplementaryContainerView.swift
+++ b/ListableUI/Sources/Internal/SupplementaryContainerView.swift
@@ -75,7 +75,7 @@ final class SupplementaryContainerView : UICollectionReusableView
     // MARK: Content
     //
     
-    weak var headerFooter : AnyPresentationHeaderFooterState? {
+    var headerFooter : AnyPresentationHeaderFooterState? {
         didSet {
             guard oldValue !== self.headerFooter else {
                 return
--- a/ListableUI/Sources/ListView/ListView.Delegate.swift
+++ b/ListableUI/Sources/ListView/ListView.Delegate.swift
@@ -163,7 +163,7 @@ extension ListView
             
             headerFooter.willDisplay(view: container)
             
-            self.displayedSupplementaryItems[ObjectIdentifier(view)] = headerFooter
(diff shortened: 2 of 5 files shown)


ListableUI/Sources/Internal/SupplementaryContainerView.swift
```diff
@@ -64,6 +64,8 @@ final class SupplementaryContainerView : UICollectionReusableView
         ) as! SupplementaryContainerView
         
         view.reuseCache = reuseCache
+        
+        // TODO: We need to also set this during on-screen updates; I don't think it's updated correctly now.
```
kyleve, line 68: To be fixed after; but found this when debugging. It's theoretical; but I think it can happen.

ListableUI/Sources/Internal/SupplementaryContainerView.swift
```diff
@@ -73,7 +75,7 @@ final class SupplementaryContainerView : UICollectionReusableView
     // MARK: Content
     //
     
-    var headerFooter : AnyPresentationHeaderFooterState? {
+    weak var headerFooter : AnyPresentationHeaderFooterState? {
```
kyleve, line 78: Fixes retain cycle on deinit that happened in some cases.

ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -163,7 +163,7 @@ extension ListView
             
             headerFooter.willDisplay(view: container)
             
-            self.displayedSupplementaryItems[ObjectIdentifier(view)] = headerFooter
+            self.displayedSupplementaryItems[ObjectIdentifier(container)] = headerFooter
```
kyleve, line 166: This is the fix! Grr.

ListableUI/Tests/Internal/SupplementaryContainerViewTests.swift
```diff
@@ -89,7 +89,9 @@ class SupplementaryContainerViewTests: XCTestCase
         
         // Add a header
         
-        view.headerFooter = self.newHeaderFooter()
+        let header = self.newHeaderFooter()
```
kyleve, line 92: Since this is now `weak`; need a strong ref to the header.

ListableUI/Sources/Internal/SupplementaryContainerView.swift
kylebshr: Is this important, or should it be a separate patch in the spirit of least change possible for the CCB?

kylebshr (commented):
LGTM - will let you decide if the non-necessary changes should go in this PR

kylebshr (approved):

ListableUI/Sources/Internal/SupplementaryContainerView.swift
kyleve: In _theory_ this is safe since the `HeaderFooterViewStatePair` retains both of these, but I can revert it out; yeah

kyleve: Removed

PR #324, opened 2021-09-03
[RetailGrid] - Pagination fixed

gabreho:


ListableUI/Sources/Layout/Retail Grid/RetailGridListLayout.swift
```diff
@@ -44,6 +44,15 @@ public struct RetailGridAppearance : ListLayoutAppearance
         public var columns: Int
         public var rows: Rows
         
+        public var paged: Bool {
```
kyleve, line 47: Minor nit: `isPaged`?

ListableUI/Sources/Layout/Retail Grid/RetailGridListLayout.swift
```diff
@@ -44,6 +44,15 @@ public struct RetailGridAppearance : ListLayoutAppearance
...
+            switch rows {
+            case .rows:
+                return true
+            default:
```
kyleve, line 51: Nit: Can you include all the cases here? That way when more cases are added; we'll get a compiler error and have to explicitly fix them. I've been bitten by `default` and `if case` too many times...

kyleve (approved):

PR #333, opened 2021-09-28
Introduce AnyHeaderFooterConvertible, so it's less verbose to create headers and footers with default values

kyleve:
### AnyHeaderFooterConvertible

This is now the type used in public APIs in place of `AnyHeaderFooter`:

```
public protocol AnyHeaderFooterConvertible {
    
    /// Converts the object into a type-erased `AnyHeaderFooter` instance.
    func toHeaderFooter() -> AnyHeaderFooter
}
```

This means that you can now do:

```swift
section.header = MyHeaderContent(title: "Albums")
```

Instead of:

```swift
section.header = HeaderFooter(MyHeaderContent(title: "Albums"))
```

In the vast majority of cases.

I've converted all the usages in tests and the demo app.

### Updated Demo Styling
Updated the demo app styling to more closely match the iOS 15 grouped table view.

### `.setting(...` functions
This introduces builder-style syntax for wrapping `ItemContent` and `HeaderFooterContent` in an `Item` and `HeaderFooter`. I don't love this name right now – looking for suggestions!

### DIFF
--- a/ListableUI/Sources/HeaderFooter/AnyHeaderFooterConvertible.swift
+++ b/ListableUI/Sources/HeaderFooter/AnyHeaderFooterConvertible.swift
@@ -1,0 +1,41 @@
+//
+//  AnyHeaderFooterConvertible.swift
+//  ListableUI
+//
+//  Created by Kyle Van Essen on 9/28/21.
+//
+
+import Foundation
+
+
+/// A type which can be converted into a `HeaderFooter`, so you
+/// do not need to explicitly wrap / convert your `HeaderFooterContent`
+/// in a `HeaderFooter` when providing an header or footer to a list or section:
+///
+/// ```
+/// Section("id") { section in
+///     section.header = MyHeaderContent(title: "Hello, World!")
+/// }
+///
+/// struct MyHeaderContent : HeaderFooterContent {
+///    var title : String
+///    ...
+/// }
+/// ```
+///
+/// Only two types conform to this protocol:
+///
+/// ### `HeaderFooter`
+/// The `HeaderFooter` conformance simply returns self.
+///
+/// ### `HeaderFooterContent`
+/// The `HeaderFooterContent` conformance returns `HeaderFooter(self)`,
+/// utilizing the default values from the `HeaderFooter` initializer.
+///
+public protocol AnyHeaderFooterConvertible {
+    
+    /// Converts the object into a type-erased `AnyHeaderFooter` instance.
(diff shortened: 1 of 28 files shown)


ListableUI/Sources/HeaderFooter/AnyHeaderFooterConvertible.swift
```diff
@@ -0,0 +1,41 @@
...
+/// The `HeaderFooter` conformance simply returns self.
+///
+/// ### `HeaderFooterContent`
+/// The `HeaderFooterContent` conformance returns `HeaderFooter(self)`,
+/// utilizing the default values from the `HeaderFooter` initializer.
+///
+public protocol AnyHeaderFooterConvertible {
+    
+    /// Converts the object into a type-erased `AnyHeaderFooter` instance.
+    func toHeaderFooter() -> AnyHeaderFooter
```
kylebshr, line 38: Doesn't really matter since this is mostly used internally, but most other types use `asAnyFoo`
```suggestion
    func asAnyHeaderFooter() -> AnyHeaderFooter
```

ListableUI/Sources/Item/Item.swift
```diff
@@ -168,6 +168,65 @@ public struct Item<Content:ItemContent> : AnyItem
...
+    ///                       sizing: .thatFits(.noConstraint),
+    ///                       selectionStyle: .tappable
+    ///                   )
+    ///
+    /// struct MyItemContent : ItemContent {
+    ///    var name : String
+    ///    ...
+    /// }
+    /// ```
+    public func setting(
```
kylebshr, line 190: What do you think about `.with`?

kylebshr (approved):

ListableUI/Sources/HeaderFooter/AnyHeaderFooterConvertible.swift
kyleve: Ahh yeah, right. Good call

ListableUI/Sources/Item/Item.swift
kyleve: Sure!

ListableUI/Sources/HeaderFooter/AnyHeaderFooterConvertible.swift
kyleve: Done

ListableUI/Sources/Item/Item.swift
kyleve: Done

PR #334, opened 2021-09-29
Prep 0.28.0

kyleve:
- Update versions to 0.28.0
- Regenerate docs

The second commit is all docs; so this commit is the one to check: https://github.com/kyleve/Listable/pull/334/commits/80cf54bcd1d0056775551d1b899dda7c606d2e0f


kylebshr (approved):

kyleve:
@tyten 

> Not being familiar with the generation here, is it fine that the paths are absolute and local?

The JSON file is only used during the generation; so it should be fine. I could likely gitignore the JSON!

PR #335, opened 2021-10-05
Allow clients to specify a `swipeActionStyle`  to customize a swipe actions view style

eugenegrebnev:
**Summary:**
Add a new property `swipeActionStyle` that allows clients to specify a style of swipe actions view (such as `standard` or `custom`). 

**Snapshot:**
swipe

### DIFF
--- a/ListableUI/Sources/Internal/DefaultSwipeView.swift
+++ b/ListableUI/Sources/Internal/DefaultSwipeView.swift
@@ -11,11 +11,47 @@ private let haptics = UIImpactFeedbackGenerator(style: .light)
 
 public final class DefaultSwipeActionsView: UIView, ItemContentSwipeActionsView {
 
+    public struct Style: Equatable {
+        public enum Shape: Equatable {
+            case rectangle(cornerRadius: CGFloat)
+        }
+
+        public static let `default` = Style()
+
+        public var actionShape: Shape
+        public var interActionSpacing: CGFloat
+        public var containerInsets: UIEdgeInsets
+
+        public init(
+            actionShape: Shape = .rectangle(cornerRadius: 0),
+            interActionSpacing: CGFloat = 0,
+            containerInsets: UIEdgeInsets = .zero
+        ) {
+            self.actionShape = actionShape
+            self.interActionSpacing = interActionSpacing
+            self.containerInsets = containerInsets
+        }
+
+        var cornerRadius: CGFloat {
+            switch actionShape {
+            case .rectangle(let cornerRadius):
+                return cornerRadius
+            }
+        }
+    }
+
     private var actionButtons: [DefaultSwipeActionButton] = []
+    private let container: UIView = {
+        let view = UIView()
+        view.backgroundColor = .clear
+        view.clipsToBounds = true
+        return view
+    }()
     private var calculatedNaturalWidth: CGFloat = 0
 
     private var firstAction: SwipeAction?
     private var didPerformAction: SwipeAction.CompletionHandler
+    private let style: Style
 
     public var swipeActionsWidth: CGFloat {
         calculatedNaturalWidth + safeAreaInsets.right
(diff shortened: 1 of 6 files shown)


ListableUI/Sources/Internal/DefaultSwipeView.swift
```diff
@@ -69,31 +116,33 @@ public final class DefaultSwipeActionsView: UIView, ItemContentSwipeActionsView
...
-        }
+        } + CGFloat(max(0, buttons.count - 1)) * style.spacing
     }
 
     public func apply(actions: SwipeActionsConfiguration) {
         if actionButtons.count != actions.actions.count {
             actionButtons.forEach { $0.superview?.removeFromSuperview() }
             actionButtons = actions.actions.map { _ in
                 let button = DefaultSwipeActionButton()
+                button.layer.cornerRadius = style.cornerRadius
```
kylebshr, line 141: Do we need to set this on the button and the `wrapperView`?

ListableUI/Sources/Internal/DefaultSwipeView.swift
```diff
@@ -11,22 +11,64 @@ private let haptics = UIImpactFeedbackGenerator(style: .light)
 
 public final class DefaultSwipeActionsView: UIView, ItemContentSwipeActionsView {
 
+    public struct Style: Equatable {
+        public enum Shape: Equatable {
+            case rectangle(cornerRadius: CGFloat)
+        }
+
+        public static let standard = Style()
```
kylebshr, line 19: I'd suggest `.default` - you can use backticks to use it even though it's a keyword
```suggestion
        public static let `default` = Style()
```

ListableUI/Sources/Internal/DefaultSwipeView.swift
```diff
@@ -11,22 +11,64 @@ private let haptics = UIImpactFeedbackGenerator(style: .light)
...
+
+        public let shape: Shape
```
kylebshr, line 21: Since some of the properties are applied to the action and some are applied to the layout, I'd suggestion more specific names:

```
public var actionShape: Shape
public var interActionSpacing: CGFloat
public var containerInsets: UIEdgeInsets
```

Also, for value types that are configurable like this I like to use vars and let the immutability of instances determine whether or not properties are mutable.

kylebshr (approved):

ListableUI/Sources/Item/ItemContent.swift
```diff
@@ -353,7 +353,10 @@ public protocol ItemContent where Coordinator.ItemContentType == Self
     /// The view type to use to render swipe actions (delete, etc) for this content.
     /// A default implementation, which matches `UITableView`, is provided.
     associatedtype SwipeActionsView: ItemContentSwipeActionsView = DefaultSwipeActionsView
-    
+
+    /// Specify a swipe action style for this content.
+    var swipeActionStyle: SwipeActionsView.Style { get }
```
kylebshr, line 358: Nit: we're styling all the actions

```suggestion
    var swipeActionsStyle: SwipeActionsView.Style { get }
```

ListableUI/Sources/Internal/DefaultSwipeView.swift
eugenegrebnev: Right, however, we can set `wrapperView.cliptoBounds = true` and do not set the `cornerRadius` for the button.

kylebshr: Oh hmm, I think I saw some weirdness at the boundaries between the views when I did that. Maybe lets leave it as is

ListableUI/Sources/Internal/DefaultSwipeView.swift
```diff
@@ -11,22 +11,64 @@ private let haptics = UIImpactFeedbackGenerator(style: .light)
 
 public final class DefaultSwipeActionsView: UIView, ItemContentSwipeActionsView {
 
+    public struct Style: Equatable {
+        public enum Shape: Equatable {
```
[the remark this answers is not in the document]
eugenegrebnev: Initially I thought that it might contain something like `circle` in future, however mirroring `Box` sounds good to me, since it's already familiar API.

eugenegrebnev:
> Nice! Mind updating the `CHANGELOG.md` at the repo root too?

Sure thing! Updated.

kyleve:
Thanks! Lmk when you want me to go ahead and merge (is now fine?)

PR #342, opened 2021-10-27
[WIP DNR] Fix List-Level Supplementary Animations

kyleve:


ListableUI/Sources/Internal/PresentationState/PresentationState.swift
```diff
@@ -113,13 +117,23 @@ final class PresentationState
     // MARK: Accessing Data
     //
     
+//    var toContent : Content {
```
kyleve, line 120: Remove

PR #352, opened 2021-11-22
Expose the natural width of a measured list for popovers, etc.

kyleve:


BlueprintUILists/Sources/List.Measurement.swift
```diff
@@ -77,7 +81,38 @@ extension List {
...
+                return horizontalFill.needsMeasurement || verticalFill.needsMeasurement
+            }
+        }
+    }
+}
+
+
+extension List.Measurement {
+    
+    public enum FillRule : Equatable {
```
kyleve, line 103: Finish docs

BlueprintUILists/Sources/List.Measurement.swift
```diff
@@ -77,7 +81,38 @@ extension List {
         ///
         case measureContent(
             cacheKey : AnyHashable? = nil,
+            horizontalFill : FillRule = .fillParent,
+            verticalFill : FillRule = .fillParent,
```
kyleve, line 85: This should be `.natural` to keep defaults

PR #367, opened 2021-12-15
ListLayout.layout(delegate:in:) now requires returning the content size, etc, to make it more obvious this is required

kyleve:
It was really easy to forget setting these properties before; so restructure the API to make them required via the return value.


ListableUI/Sources/Layout/ListLayout/ListLayoutValues.swift
```diff
@@ -1,56 +0,0 @@
...
-//  ListableUI
-//
-//  Created by Kyle Van Essen on 6/6/20.
-//
-
-import Foundation
-import UIKit
-
-
-public struct ListLayoutPoint : Hashable {
```
kyleve, line 12: Housekeeping, these were no longer used.

n8chur (approved):

AdamPro13 (approved):

PR #372, opened 2022-01-15
A swipe actions memory leak has been fixed for ItemCell.ContentContainerView

eugenegrebnev:
During a memory leaks investigation we found that there is a leak in `ItemCell.ContentContainerView`. 

In `registerSwipeActionsIfNeeded` function, we create a `Content.SwipeActionsView` instance and pass a `self.didPerformAction` function that captures `self`. And since the `Content.SwipeActionsView` instance is added to `self` as a subview, we have a retain cycle.

The changes here break the cycle.

### DIFF
--- a/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -102,7 +102,12 @@ extension ItemCell {
         public func registerSwipeActionsIfNeeded(actions: SwipeActionsConfiguration, style: Content.SwipeActionsView.Style, reason: ApplyReason) {
             if swipeConfiguration == nil {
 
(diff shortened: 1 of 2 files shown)


ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -102,7 +102,12 @@ extension ItemCell {
         public func registerSwipeActionsIfNeeded(actions: SwipeActionsConfiguration, style: Content.SwipeActionsView.Style, reason: ApplyReason) {
             if swipeConfiguration == nil {
 
-                let swipeView = Content.SwipeActionsView(style: style, didPerformAction: self.didPerformAction)
+                let swipeView = Content.SwipeActionsView(
```
kyleve, line 105: Oooh, very good catch!

kyleve (approved):

PR #374, opened 2022-02-04
Reordering items with long press support

gabreho:

- `[ItemReordering.GestureRecognizer` - Now it is a `UILongPressGestureRecognizer`. Instead of reporting that the drag event started in the update method, we now have `gestureRecognizerShouldBegin` so we can catch exactly when the press duration requirement has passed.
- `[ListReorderGesture` - It now has `Begins`, an enum that configures the `GestureRecognizer`'s `minimumPressDuration`.

### Checklist

### DIFF
--- a/BlueprintUILists/Sources/ListReorderGesture.swift
+++ b/BlueprintUILists/Sources/ListReorderGesture.swift
@@ -38,14 +38,19 @@ import UIKit
 /// ```
 public struct ListReorderGesture : Element
 {
+    public enum Begins {
+        case onTap
+        case onLongPress
+    }
+
     /// The element which is being wrapped by the reorder gesture.
     public var element : Element
     
     /// If the gesture is enabled or not.
     public var isEnabled : Bool
 
-    /// TODO
-    public var requiresLongPress: Bool
+    /// Condition to start the reorder gesture
+    public var begins: Begins
     
     let actions : ReorderingActions
     
@@ -56,14 +61,14 @@ public struct ListReorderGesture : Element
     public init(
         isEnabled : Bool = true,
         actions : ReorderingActions,
-        requiresLongPress: Bool = false,
+        begins: Begins = .onTap,
         wrapping element : Element
     ) {
         self.isEnabled =  isEnabled
         
         self.actions = actions
 
-        self.requiresLongPress = requiresLongPress
+        self.begins = begins
         
         self.element = element
     }
@@ -88,7 +93,7 @@ public struct ListReorderGesture : Element
                 
                 view.recognizer.apply(actions: self.actions)
                 
-                view.recognizer.minimumPressDuration = requiresLongPress ? 0.5 : 0.0
+                view.recognizer.minimumPressDuration = begins == .onLongPress ? 0.5 : 0.0
             }
         }
     }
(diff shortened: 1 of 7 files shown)


ListableUI/Sources/Item/ItemReordering.swift
```diff
@@ -182,51 +181,58 @@ extension ItemReordering {
...
             @unknown default: listableInternalFatal()
             }
         }
+
+        private var initialTouchPoint : CGPoint? = nil
+
+        public func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+            let shouldStart = onStart?() ?? false
+            if shouldStart {
+                initialTouchPoint = location(in: view)
```
kyleve, line 232: Fwiw, you shouldn't store state in `gestureRecognizerShouldBegin` – the system can still cancel the gesture after this happens. This caused bugs in Market until recently when we resolved it. You should respond to the gesture in its target/action callbacks.

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -43,6 +43,9 @@ public struct ListReorderGesture : Element
     
     /// If the gesture is enabled or not.
     public var isEnabled : Bool
+
+    /// TODO
+    public var requiresLongPress: Bool
```
kyleve, line 48: I think this might be nicer as an enum, eg:

```
enum Begins {
  case onTap
  case onLongPress
}
```

So then the syntax is like:

```
myElement
   .listReorderGesture(with actions, begins: .onLongPress)
```

Demo/Sources/Demos/Demo Screens/CollectionViewAppearance.swift
```diff
@@ -132,10 +132,16 @@ struct DemoHeader2 : BlueprintHeaderFooterContent, Equatable
...
 {
     var text : String
+    var dragging = false
+    var requiresLongPress = false
     
     var identifierValue: String {
         return self.text
     }
+    
+    func isEquivalent(to other: DemoItem) -> Bool {
```
kyleve, line 142: Nit: Not needed I don't think.

Demo/Sources/Demos/Demo Screens/CollectionViewAppearance.swift
```diff
@@ -132,10 +132,16 @@ struct DemoHeader2 : BlueprintHeaderFooterContent, Equatable
 struct DemoItem : BlueprintItemContent, Equatable, LocalizedCollatableItemContent
 {
     var text : String
+    var dragging = false
```
kyleve, line 135: Do you need this? This is on the ItemState in ApplyItemContentInfo https://github.com/kyleve/Listable/blob/main/ListableUI/Sources/Item/ItemState.swift#L35

gabreho:
@kyleve this looks way cleaner after your suggestions.

ListableUI/Sources/Item/ItemReordering.swift
```diff
@@ -167,9 +167,7 @@ extension ItemReordering {
             super.init(target: target, action: action)
             
             self.addTarget(self, action: #selector(updated))
-            
-            self.minimumNumberOfTouches = 1
-            self.maximumNumberOfTouches = 1
+            self.minimumPressDuration = 0.5
```
kyleve, line 170: Extremely minor nit: I'd leave this as 0, or pass the time through in the init method as well.

kyleve: (Not that it matters since it gets applied by the `ViewDescription` anyway for our purposes – but if you're not using the Blueprint backed API, passing in the `Begins` likely makes sense so we can provide the same time values)

kyleve (approved):

kyleve:
Looks good; one optional thing. Thanks!

ListableUI/Sources/Item/ItemReordering.swift
gabreho: makes sense. Changed to `0`

PR #376, opened 2022-02-15
reordering now possible with voiceover enabled

RoyalPineapple:
Configured Listable to support accessibility reordering. 

https://user-images.githubusercontent.com/219578/154136135-6463c2bf-3d23-4b5c-ae8b-43831bda35dc.mov

### Checklist


BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -138,3 +153,15 @@ fileprivate extension ListReorderGesture
...
 }
+
+private class AccessibilityReorderAction: UIAccessibilityCustomAction {
+    
+}
+
+fileprivate extension ListReorderGesture
+{
+    struct Strings{
+        static var accessibilityLabel = NSLocalizedString("Reorder", comment: "Accessibility label for the reorder control on an item")
```
RoyalPineapple, line 164: I think this is the first localized string in this repo. We might prefer to have these passed in at load rather than set up localization.

Demo/Sources/Demos/Demo Screens/CollectionViewAppearance.swift
```diff
@@ -163,7 +163,12 @@ struct DemoItem : BlueprintItemContent, Equatable, LocalizedCollatableItemConten
             }
         }
         .inset(horizontal: 15.0, vertical: 13.0)
-        .accessibilityElement(label: self.text, value: nil, traits: [.button])
+        
+        if !info.isReorderable {
+            // Only wrap the row in an accessibility element if we arent showing the reorder gesture.
+            return AccessibilityElement(label: self.text, value: nil, traits: [.button], wrapping: row)
```
RoyalPineapple, line 169: This is going to be hard to enforce externally.

BlueprintUILists/Sources/ListReorderGesture.swift
kyleve: Yeah, these will need to be passed in since we don't have the ability to wire this up to Shuttle. We could inject this in in Market, which does have loc, I think.

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -138,3 +153,15 @@ fileprivate extension ListReorderGesture
         }
     }
 }
+
+private class AccessibilityReorderAction: UIAccessibilityCustomAction {
```
kyleve, line 157: Remove?

ListableUI/Sources/Internal/ItemCell.swift
```diff
@@ -161,3 +188,26 @@ final class LiveCells {
...
+extension UIView {
+    
+   fileprivate func firstAccessibleChild() -> UIView? {
+        guard !isAccessibilityElement else {
+            return self
+        }
+        return recursiveAccessibleSubviews().first as? UIView
+    }
+    
+    fileprivate func recursiveAccessibleSubviews() -> [Any] {
```
kyleve, line 202: Can this be a `UIView`, not `Any`?

BlueprintUILists/Sources/ListReorderGesture.swift
RoyalPineapple: oops, that should be in a different branch

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1333,6 +1333,54 @@ extension ListView : ReorderingActionsDelegate
...
+                }
+            } else {
+                destinationPath = IndexPath(row: indexPath.row + 1, section: indexPath.section)
+            }
+        }
+        
+        let targetPath =  self.delegate.collectionView(self.collectionView, targetIndexPathForMoveFromItemAt: indexPath, toProposedIndexPath: destinationPath)
+        
+        item.beginReorder(from: indexPath, with: self.environment)
+        self.dataSource.collectionView(self.collectionView, moveItemAt: indexPath, to: targetPath)
```
RoyalPineapple, line 1377: I feel like this should be called by the collection view itself when we call `moveItem(at: indexPath, to: targetPath)` below, but it doesn't seem to.

BlueprintUILists/Resources/en.lproj/Localizable.strings
```diff
@@ -0,0 +1,12 @@
+/* Accessibility hint for the reorder control in an item */
+"Double tap and hold, wait for the sound, then drag to rearrange." = "Double tap and hold, wait for the sound, then drag to rearrange.";
```
kyleve, line 2: Should this be in BlueprintUILists? I would expect this to be in ListableUI.

kyleve: (Eg, if / when we support SwiftUI, ideally we can reuse the same strings. Or the strings can be used by Listable without Blueprint)

Demo/Sources/Demos/Demo Screens/CollectionViewAppearance.swift
```diff
@@ -163,7 +163,12 @@ struct DemoItem : BlueprintItemContent, Equatable, LocalizedCollatableItemConten
             }
         }
         .inset(horizontal: 15.0, vertical: 13.0)
-        .accessibilityElement(label: self.text, value: nil, traits: [.button])
+        
+        if !info.isReorderable {
+            // Only wrap the row in an accessibility element if we arent showing the reorder gesture.
```
kyleve, line 168: Is this something consumers will have to do manually as well?

ListableUI/Sources/Internal/ItemCell.swift
```diff
@@ -136,6 +141,28 @@ final class ItemCell<Content:ItemContent> : UICollectionViewCell, AnyItemCell
...
     }
+    
+    
+    // MARK: AccessibilityLabel
+    
+    // When reordering cells the UICollectionView expects all cells to have a valid accessibility label, even when acting as an accessibility container with `isAccessibilityElement == false`. This is used to announce the destination of the reodering operaton in relation to the other cells, e.g. "Before foo" or "after bar".
+    private var _accessibilityLabel: String?
+    override var accessibilityLabel: String? {
+        set {
+            guard _accessibilityLabel != newValue  else {
```
kyleve, line 152: Nit: This guard isn't very useful, since it's not protecting any expensive or non-idempotent work... just remove?

ListableUI/Sources/Internal/ItemCell.swift
```diff
@@ -136,6 +141,28 @@ final class ItemCell<Content:ItemContent> : UICollectionViewCell, AnyItemCell
...
+                return
+            }
+            _accessibilityLabel = newValue
+        }
+        get {
+            guard let accessibilityLabel = _accessibilityLabel else {
+                let first = contentView.firstAccessibleChild()?.accessibilityLabel
+                _accessibilityLabel = first
```
kyleve, line 160: Nit: Getters shouldn't be performing side effects. Can we just return return `first` without setting this value?

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1333,6 +1333,56 @@ extension ListView : ReorderingActionsDelegate
         }
     }
     
+    func accessibilityMove(item: AnyPresentationItemState, direction: ReorderingActions.AccessibilityMoveDirection) -> Bool {
```
kyleve, line 1336: Can you comment / document what we need this for / what we lose by not doing it? I am a bit nervous about invoking these delegate and data source methods ourselves...

BlueprintUILists/Resources/en.lproj/Localizable.strings
RoyalPineapple: the gesture recognizer where it gets used is within `BlueprintUILists` so i just matched that. I can relocate it.

Demo/Sources/Demos/Demo Screens/CollectionViewAppearance.swift
RoyalPineapple: if the call is wrapped in an `AccessibilityElement`, yes.  The view containing the gesture recognizer needs to be reachable by voiceover so its containing view must be acting as a container.

ListableUI/Sources/Internal/ItemCell.swift
RoyalPineapple: 👍🏻

ListableUI/Sources/Internal/ItemCell.swift
```diff
@@ -161,3 +183,26 @@ final class LiveCells {
...
 }
+
+
+extension UIView {
+    
+   fileprivate func firstAccessibleChild() -> NSObject? {
+        guard !isAccessibilityElement else {
+            return self
+        }
+        return recursiveAccessibleSubviews().first as? NSObject
```
kyleve, line 194: Can we make `recursiveAccessibleSubviews` also return `NSObject` so we don't need that check here?

ListableUI/Sources/ListView/ListView.swift
kyleve: Mind enumerating for the reviewers?

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1333,6 +1333,61 @@ extension ListView : ReorderingActionsDelegate
...
+        guard let indexPath = self.storage.presentationState.indexPath(for: item),
+              self.dataSource.collectionView(self.collectionView, canMoveItemAt: indexPath) else {
```
kyleve, line 1338: Spacing is off here

ListableUI/Sources/ListView/ListView.swift
kyleve: Ping on this

ListableUI/Sources/LocalizedStrings.swift
```diff
@@ -0,0 +1,43 @@
...
+//  LocalizedStrings.swift
+//  BlueprintUILists
+//
+//  Created by Alex Odawa on 22/02/2022.
+//
+
+import Foundation
+
+
+public struct LocalizedStrings {
```
kyleve, line 11: Nit: This is public, so anyone who imports this will get this as a top level name – and its a common name, meaning conflicts requiring folks to fully qualify. Can we rename to `ListableLocalizedStrings`?

ListableUI/Sources/LocalizedStrings.swift
```diff
@@ -0,0 +1,43 @@
...
+    
+    public struct ReorderGesture{
+ 
+
+        public static var accessibilityLabel = NSLocalizedString("reorder.AccesslibilityLabel",
```
kyleve, line 16: Nit: `let`, otherwise anyone can now assign this value which is not what we want.

kyleve: (Same for other properties in here)

kyleve: Also, "Accesslibility" is mis-spelled; should be "Accessibility"

ListableUI/Sources/ReorderingActions.swift
```diff
@@ -79,10 +87,19 @@ extension ReorderingActions {
     }
 }
 
+extension ReorderingActions {
+    public enum AccessibilityMoveDirection {
```
kyleve, line 91: Nit: Docs

ListableUI/Sources/ListView/ListView.swift
RoyalPineapple: Added a comment explaining in the protocol definition. 

>    In addition to reordering cells with the standard drag gesture we offer an AccessibilityCustomAction to move a selected cell up or down by a single index path position. This provides an affordance for those who struggle to precisely drag cells about the screen.

ListableUI/Sources/ReorderingActions.swift
RoyalPineapple: added a comment

kyleve (approved):
Looks good! One other small ask: Can you update RELEASING.md to mention the strings process?

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1333,6 +1333,61 @@ extension ListView : ReorderingActionsDelegate
...
+                }
+            } else {
+                destinationPath = IndexPath(row: indexPath.row - 1, section: indexPath.section)
+            }
+
+        case .down:
+            // Moving an item down means incrementing the index.
+            if indexPath.row == storage.allContent.sections[indexPath.section].count - 1 {
+                // we are the last item our section, lets see if there's another section we can move down to
+                if storage.allContent.sections.count > indexPath.section {
```
kylebshr, line 1365: Maybe I'm wrong, but don't we need to decrement the sections.count by 1 for this comparison, since `indexPath.section` is zero indexed? e.g., if there's three sections, and we're on the last section, the section will be 2, the count will be 3, so `count > section = true` but we can't increase the section past 2

```suggestion
                if storage.allContent.sections.count - 1 > indexPath.section {
```

kylebshr (approved):

ListableUI/Sources/ListView/ListView.swift
RoyalPineapple: Yep, of course.

PR #378, opened 2022-03-21
Fixed list measurements with container headers

alexisakers:
When measuring the content size in `ListView.contentSize`, the height of large titles would be omitted. This happened because the measuring presentation state was using the `header` instead of the `containerHeader` when building the state for measuring.

| Before | After |
|---|---|
image | Simulator Screen Shot - SQ iPad 14 5 - 2022-03-21 at 11 39 28 |

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # [Main]
 
 ### Fixed
-- Fix measuring large titles in modals.
+- Fixed list measurements with container headers.
 
 ### Added
 - Item reordering is now possible when using VoiceOver.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # [Main]
 
 ### Fixed
+- Fix measuring large titles in modals.
 
 ### Added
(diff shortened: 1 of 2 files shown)


alexisakers:
It looks like the test failures could be related to the Apple outage, causing a download issue. Might re-run the tests after that clears up to double check.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # [Main]
 
 ### Fixed
+- Fix measuring large titles in modals.
```
kyleve, line 4: Nit: large titles or modals aren't a thing here. Can you instead say something like `Fixed list measurements with container headers"?

kyleve (approved):

kyleve:
> It looks like the test failures could be related to the Apple outage, causing a download issue. Might re-run the tests after that clears up to double check.

If the 15 and 14 one pass, we should be good to go here!

CHANGELOG.md
alexisakers: Oops that's right, updated

PR #379, opened 2022-03-21
Prep 3.2.0

kyleve:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,8 @@
 # [Main]
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


kylebshr:
This won't make it in the branch cut unless it's 3.1.1 right? We probably want to relax the Market requirement

kyleve:
Hmm, you're right. I thought we'd changed the Market requirement to be ~> 3.0, but I guess not... This does include new features (the a11y reordering work Alex did), so 3.2.0 is right here.

kylebshr (approved):

PR #381, opened 2022-04-01
Don't force inherit animations when updating blueprint content.

kyleve:
This doesn't seem to be needed anymore. Animations still occur, I suspect later changes to unify layout passes fixed this.

### Checklist


n8chur (approved):
I guess this seems like it should work? Approving assuming you'll test this integrated in a few places in Register/Market :+1:

PR #387, opened 2022-05-31
Add pin-scrolling behavior

gilgtm:
- Adds a pinning feature which is very similar to scrollToItem except that you don't specify an onInsertOf item.
- Adds a bottomScrollOffset property to ListScrollPositionInfo. You can use this to fine-tune pinning by only pinning when within a certain distance of the bottom scroll position

### Checklist

### DIFF
--- a/ListableUI/Sources/ListScrollPositionInfo.swift
+++ b/ListableUI/Sources/ListScrollPositionInfo.swift
@@ -29,6 +29,9 @@ public struct ListScrollPositionInfo : Equatable {
     
     /// If the last item list is partially visible.
     public var isLastItemVisible : Bool
+
+    /// Distance required to scroll to the bottom
+    public var bottomScrollOffset: CGFloat
     
     ///
     /// Used to retrieve the visible content edges for the list's content.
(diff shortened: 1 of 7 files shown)


ListableUI/Sources/ListScrollPositionInfo.swift
```diff
@@ -106,6 +109,8 @@ public struct ListScrollPositionInfo : Equatable {
         
         self.isFirstItemVisible = isFirstItemVisible
         self.isLastItemVisible = isLastItemVisible
+
+        self.bottomScrollOffset = scrollView.contentSize.height - scrollView.bounds.size.height - scrollView.contentOffset.y
```
kyleve, line 113: Any idea if we need to include some safe area calc here? (Sorry, I forget how these properties all relate to each other...)

kyleve (approved):
Nice, thank you! I can get a release out once this is merged

ListableUI/Sources/ListScrollPositionInfo.swift
gilgtm: good callout, we need to consider the adjusted insets, not just the safe area insets

PR #391, opened 2022-06-24
Ensure we properly fade supplementary views in and out

kyleve:
Because we don't actually remove the real supplementary views (they're always there – we have to do this because there's no dynamic way to add or remove them without reloading the entire section) and just update the content of the real supplementary views, the standard animation for these was never occurring. Implement a simple fade for now. Eventually we could pass through a customizable animation like we do for `Item` insertions and removals.

### Checklist


ListableUI/Sources/Internal/SupplementaryContainerView.swift
```diff
@@ -107,20 +113,51 @@ final class SupplementaryContainerView : UICollectionReusableView
     var environment : ListEnvironment!
     var reuseCache : ReusableViewCache!
     
-    private(set) var content : UIView? {
-        didSet {
-            if let old = oldValue {
+    func setContent(_ new : UIView?, animated: Bool) {
```
kyleve, line 116: Important bit is here!

ListableUI/Sources/Internal/SupplementaryContainerView.swift
```diff
@@ -107,20 +113,51 @@ final class SupplementaryContainerView : UICollectionReusableView
...
+            } else {
                 old.removeFromSuperview()
             }
-            
-            if let new = self.content {
+        }
+        
+        if let new = self.content {
+            if animated {
+                UIView.performWithoutAnimation {
```
nononoah, line 142: Is this to ensure that we don't animate the addition in the case where this is invoked as part of a parent animation?

nononoah (approved):

ListableUI/Sources/Internal/SupplementaryContainerView.swift
kyleve: Yessir

PR #395, opened 2022-07-14
Include Safe Area in Measurement

kyleve:
This updates list measurement to properly include the safe area from the environment – that way this https://github.com/squareup/market/pull/4237 sort of case will work correctly.

### Checklist


ListableUI/Sources/Layout/ListLayout/ListLayoutContent.swift
```diff
@@ -237,8 +237,12 @@ public final class ListLayoutContent
     // MARK: Performing Layouts
     //
     
-    func apply(result : ListLayoutResult) {
-        self.contentSize = result.contentSize
+    func apply(result : ListLayoutResult, adjustedContentInset: UIEdgeInsets) {
```
kyleve, line 240: Move this to measurement only. We don't want this in here always because the scroll view will handle it based on the `contentAdjustmentBehavior`

nononoah (approved):

kylebshr:
I'm not entirely sure this makes sense; the issue I had in Blueprint (and I think the same issue is here) is that the safe area insets from the environment aren't always accurate/aren't always applied. 

- If the scroll view has explicit insets that are larger than the safe area, they're not applied
- If the scroll view is inset from the blueprint view, they won't be applied

I think we need to use the adjustedContentInset to solve this, maybe that's possible here but it's not in Blueprint since we don't have the backing view at measurement time.

kyleve:
> I'm not entirely sure this works correctly; the issue I had in Blueprint (and I think the same issue is here) is that the safe area insets from the environment aren't always accurate/aren't always applied.
> 
> * If the scroll view has explicit insets that are larger than the safe area, they're not applied
> * If the scroll view is inset from the blueprint view, they won't be applied
> 
> I think we need to read the adjustedContentInset to solve this, maybe that's possible here but it's not in Blueprint since we don't have the backing view at measurement time.

Yeah you'd need to adjust the safe area from the environment if the safe area differs from the SA your view will be presented. It seems to work fine though; perhaps maybe you should have to explicitly pass in a safe area to make this more clear.

kylebshr:
We can try it, I think an integration test in Market & Register is worth it before merging though

BlueprintUILists/Sources/List.Measurement.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import ListableUI
+import SwiftUI
```
kyleve, line 9: Grr Xcodeeee

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Changed
 
+- `List` measurement now includes the `safeAreaInsets` from the `Environment` if the `contentInsetAdjustmentBehavior` will affect the layout.
```
kylebshr, line 11:
```suggestion
- `List` measurement now has options to include the `safeAreaInsets` from the `Environment` if the `contentInsetAdjustmentBehavior` will affect the layout.
```

kylebshr (approved):

PR #400, opened 2022-07-19
Ensure we respect defaults values

kyleve:
Fixes an issue where `Optional` default values would get stamped, because passing them through to the func ends up doubly wrapping them to `Optional?` (aka `Value??`, of which the inner one is always `.some(nil)`.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@
 
 # Past Releases
 
+# [5.0.1] - 2022-07-19
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


kylebshr (approved):
Nesting `??` can be super slow to compile (or used to be at least), but if it seemed fine it's fine with me!

kyleve:
> Nesting `??` can be super slow to compile (or used to be at least), but if it seemed fine it's fine with me!

It _seems_ OK here since we already have a defined type we're assigning to, I think

PR #410, opened 2022-07-22
Add accessibility parameters to SwipeAction

kyleve:
Some places in POS are passing no title, which means a11y won't work in those cases – add options for various a11y strings.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


RoyalPineapple (approved):
Ah, good catch. 
I like the pattern of requiring either a title or accessibilityLabel, rather than just setting the accessibility label to the title by default.

mattfaluotico (approved):

PR #411, opened 2022-07-23
Use item insert/removal animations for sections with one item and no header + footer

kyleve:
There's a somewhat common pattern to add an empty state, some other banner, etc, in a `Section` with a single row, and no header or footer. In these cases, the section is basically "invisible", the point is showing the row's content. Unfortunately though, because the `Section` is the thing being inserted here, _not_ the single row of content, the standard section insertion animation is used, instead of the item insertion animation.

This PR changes the logic of what animation to use, so if a section is inserted that has: A single item, no header, and no footer, the insertion animation for the single item will be used instead.

### Checklist


kylebshr:
I won't fight you too hard on this, but it doesn't feel like something that should be implemented at the Library level - if you want this, you should just have an empty section always there and conditionally add/remove the element.

kyleve:
> I won't fight you too hard on this, but it doesn't feel like something that should be implemented at the Library level - if you want this, you should just have an empty section always there and conditionally add/remove the element.

I went back and forth on this a few times, but landed here for a couple reasons:

1. I've seen this question come up from time to time – eg, insert animations aren't "working" for this use case; which I recently realized was because of the behavior this fixes.

2. Re this:

> if you want this, you should just have an empty section always there and conditionally add/remove the element

This isn't always doable or realistic: Many list configs specify spacing between sections; so you'd end up with blank space in the list (I suppose we could change behavior to filter out completely empty sections, which may be a good change regardless). But requiring an empty section feels somewhat counter-intuitive to me / a bit of a gotcha – whereas IMO this change makes the list do what you'd intuitively "expect".

kylebshr (approved):

kyleve:
Reverting, caused some crashes: https://github.com/kyleve/Listable/pull/431

PR #412, opened 2022-07-25
[WIP DNR] Improve using elements in lists

kyleve:

### Checklist


ListableUI/Sources/Section/Section.swift
```diff
@@ -138,24 +138,6 @@ public struct Section
         self.footer = footer()
     }
     
-    /// Creates a new section with result builder-style APIs.
-    public init<IdentifierValue:Hashable>(
```
kyleve, line 142: This was a pure dupe of the above, just with less parameters.

ListableUI/Sources/ListableBuilder.swift
```diff
@@ -86,3 +86,52 @@
...
+/// }
+/// ```
+///
+/// ## Links & Videos
+/// https://github.com/apple/swift-evolution/blob/main/proposals/0289-result-builders.md
+/// https://developer.apple.com/videos/play/wwdc2021/10253/
+/// https://www.swiftbysundell.com/articles/deep-dive-into-swift-function-builders/
+/// https://www.avanderlee.com/swift/result-builders/
+///
+@resultBuilder public enum ListableOptionalBuilder<ContentType> {
```
kyleve, line 108: Perhaps remove this for now. It's tricky to use right

BlueprintUILists/Sources/Element+HeaderFooter.swift
```diff
@@ -0,0 +1,89 @@
...
+        HeaderFooter(
+            WrappedHeaderFooterContent(
+                represented: self
+            ),
+            configure: configure
+        )
+    }
+    
+    /// Used by internal Listable methods to  convert type-erased `Element` instances into `Item` instances.
+    func toHeaderFooterConvertible() -> AnyHeaderFooterConvertible {
```
kyleve, line 45: Todo: Consider using generics at the result builder level, instead of just `Element`, so the `where ElementType : Equatable`, etc, extensions work correctly.

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -0,0 +1,122 @@
...
+            WrappedElementContent(
+                represented: self,
+                identifierValue: id
+            ),
+            configure: configure
+        )
+    }
+    
+    /// Used by internal Listable methods to  convert type-erased `Element` instances into `Item` instances.
+    func toAnyItemConvertible() -> AnyItemConvertible {
```
kyleve, line 75: Todo: Consider using generics at the result builder level, instead of just Element, so the where ElementType : Equatable, etc, extensions work correctly.

BlueprintUILists/Sources/Section+Element.swift
```diff
@@ -0,0 +1,37 @@
...
+//  BlueprintUILists
+//
+//  Created by Kyle Van Essen on 7/24/22.
+//
+
+import BlueprintUI
+import ListableUI
+
+
+extension Section {
```
kyleve, line 12: Is this useful at all? Consider just removing.

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -0,0 +1,137 @@
...
+    ///     }
+    /// ```
+    ///
+    /// ## ⚠️ Performance Considerations
+    /// Unless your `Element` conforms to `Equatable` or `IsEquivalentContent`,
+    /// it will return `false` for `isEquivalent` for each content update, which can dramatically
+    /// hurt performance for longer lists (eg, more than 20 items): it will be re-measured for each content update.
+    ///
+    /// It is encouraged for these longer lists, you ensure your `Element` conforms to one of these protocols.
+    public func item(
```
kylebshr, line 33: Can I suggest `listItem`, since this will show up for anyone import Listable and `item` is fairly amiguous

```suggestion
    public func listItem(
```

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -0,0 +1,137 @@
...
+        id : AnyHashable = ObjectIdentifier(Self.Type.self),
```
kylebshr, line 34: Is this ID useful? I imagine most elements in a list will be the same type

kylebshr (approved):
I think some discussion points here would be:

a) Is this an even bigger foot gun than what we have now? Many people probably won't make their element equatable or `IsEquivalentItem`
b) Is there a performance concern from losing identifiers (most people won't add one)
c) Most elements have closures, which make equatability impossible. What's the correct thing to do there. Should there be more helpers for interacting with the row tap handlers, so that the element doesn't have to be re-applied for the closure to be correct?

ListableUI/Sources/IsEquivalent/IsEquivalentContent.swift
```diff
@@ -0,0 +1,91 @@
...
+
+import Foundation
+
+
+/// Used by the list to determine when the content of content has changed; in order to
+/// remeasure the content and re-layout the list.
+///
+/// Please see ``IsEquivalentContent/isEquivalent(to:)-15tcq`` for a full discussion of
+/// correct (and incorrect) implementation and usages.
+public protocol IsEquivalentContent {
```
kylebshr, line 16: `ContentComparable`? `EquivalentContent`? The name certainly shouldn't begin with `Is`

kyleve: `EquivalentComparable` or `EquivalentContent` seem best to me?

BlueprintUILists/Sources/Element+Item.swift
kyleve: Maybe, but not always! I might just make the default nil, because the list under the hood will already splice IDs by type anyway.

BlueprintUILists/Sources/Element+Item.swift
kyleve: Good point, agreed. I'll change the other one to `listHeaderFooter` to be consistent too?

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -0,0 +1,137 @@
+//
+//  Element+Item.swift
```
kyleve, line 2: @kylebshr Creating a thread to reply to stuff, so we can talk in a thread vs top level comments:

> a) Is this an even bigger foot gun than what we have now? Many people probably won't make their element equatable or IsEquivalentItem

Hard to say – there's a bunch of existing `ElementItem` usages in POS now, and they're all pretty much wrong – the API was hard to use, so there's a lot of this:

```
ElementItem("formFields-element", id: \.self) { _, _ in
   // Return an element
}

ElementItem(transaction.token, id: \.self) { _, _ in ... }

ElementItem("error-state", id: \.self) { _, _ in ... }

ElementItem("business-organization-name-content", id: \.self) { _, _ in ... }

Etc...
```

But in this case the underlying value is `"formFields-element"`, aka a static string; so `isEquivalent` always returns true, and the item will never re-size even when it should – which is already a foot gun and breaks things in weird ways.

> b) Is there a performance concern from losing identifiers (most people won't add one)

Identifiers? No – the list smart enough to do a "best attempt" at creating stable identifiers when there's duplicate IDs (and identifiers are already salted with the `ItemContent` type), so even without explicitly provided IDs, you'll get stable IDs across updates. Eg, if this is your list, these will be the identifiers:

```
MyElement1() -> Identifier(MyElement1, 1)
MyElement2() -> Identifier(MyElement2, 1)
MyElement2() -> Identifier(MyElement2, 2)
MyElement1() -> Identifier(MyElement1, 2)
```

The main benefit to providing IDs is during mutative diffs, the list can more intelligently manage the changes.

> c) Most elements have closures, which make equatability impossible. What's the correct thing to do there. Should there be more helpers for interacting with the row tap handlers, so that the element doesn't have to be re-applied for the closure to be correct?

Same thing as you'd do before with `ItemContent` / `BlueprintItemContent`; you need to manually implement `Equatable` or `isEquivalent`, and compare the equatable parameters that you can. There are some clever things we can do here to avoid needing this, but I didn't explore it: We can use `mirror` to find equatable parameters and compare them

kyleve: Was able to resolve (a)! We can do this automagically now. Check out `areEquatablePropertiesEqual`: https://github.com/kyleve/Listable/pull/412/files#diff-5dc21cc38bacb2f48c0751e2b61118dc52d866ca8333c61bc050d459bba4a35f

BlueprintUILists/Sources/Element+Item.swift
kyleve: Done!

BlueprintUILists/Sources/Element+Item.swift
kyleve: I made this nil

ListableUI/Sources/IsEquivalent/IsEquivalentContent.swift
kyleve: We could also go with literally just `IsEquivalent`? `Equivalentable`? (ehhh probably not)

ListableUI/Sources/IsEquivalent/CompareEquatableProperties.swift
```diff
@@ -0,0 +1,133 @@
...
+        }
+        
+        // 3b) Finally, compare the underlying values.
+        
+        guard isEqual(prop1.value, prop2.value) else {
+            return false
+        }
+    }
+    
+    // 4) All `Equatable` properties were equal, so we're equal.
```
kyleve, line 68: Wonder if we should also handle (a) one or more properties, (b) none of them are equatable. I think this means we might want to always return false.

CHANGELOG.md
```diff
@@ -6,10 +6,31 @@
...
+                item.insertAndRemoveAnimations = .scaleUp
+            }
+    }
+    ```
+    
+- Added `ListableValueBuilder`, a result builder for single-value results.
+
 ### Removed
 
 ### Changed
```
kyleve, line 28: - [ ] Missing: Default `isEquivalent` implementation.

ListableUI/Sources/IsEquivalent/AlwaysEqual.swift
```diff
@@ -0,0 +1,22 @@
...
+//  AlwaysEqual.swift
+//  ListableUI
+//
+//  Created by Kyle Van Essen on 7/27/22.
+//
+
+import Foundation
+
+
+@propertyWrapper public struct AlwaysEqual<Value> : Equatable {
```
kyleve, line 11: Remove

ListableUI/Sources/IsEquivalent/IsEquivalentContent.swift
kyleve: Went with `EquivalentComparable`

ListableUI/Sources/IsEquivalent/CompareProperties.swift
```diff
@@ -0,0 +1,249 @@
...
+///     var detail : String?
+///     var count : Int
+///
+///     var nonEquatable: NonEquatableValue
+///
+///     var closure : () -> ()
+/// }
+/// ```
+///
+/// Inspired by https://github.com/objcio/S01E264-comparing-views/blob/master/Sources/NotSwiftUIState/AnyEquatable.swift
```
kyleve, line 31: Also link to https://talk.objc.io/episodes/S01E264-comparing-views

BlueprintUILists/Sources/ListableBuilder+Element.swift
```diff
@@ -0,0 +1,87 @@
...
+/// Section("3") { section in
+///     TestContent1() // An ItemContent
+///
+///     Element1() // An Element
+///     Element2() // An Element
+/// }
+/// ```
+///
+/// ## Note
+/// Takes advantage of `@_disfavoredOverload` to avoid ambiguous method resolution with the default implementations.
```
kyleve, line 24: Comment is out of date

BlueprintUILists/Tests/Element+ItemTests.swift
```diff
@@ -0,0 +1,17 @@
...
+//  BlueprintUILists-Unit-Tests
+//
+//  Created by Kyle Van Essen on 7/27/22.
+//
+
+import XCTest
+import BlueprintUILists
+
+
+class Element_ItemTests : XCTestCase {
```
kyleve, line 12: Empty file

ListableUI/Sources/IsEquivalent/CompareProperties.swift
```diff
@@ -0,0 +1,249 @@
...
+///
+///     var title : String
+///     var detail : String?
+///     var count : Int
+///
+///     var nonEquatable: NonEquatableValue
+///
+///     var closure : () -> ()
+/// }
+/// ```
```
kyleve, line 29: Also note that Equatable objects will be compared without the traversal.

ListableUI/Sources/IsEquivalent/CompareProperties.swift
```diff
@@ -0,0 +1,249 @@
...
+        return .hasNoFields
+    }
+    
+    let rhs = Mirror(reflecting: rhs)
+    
+    // Enumerate each property by enumerating the value's `Mirror`.
+    
+    var hadEquatableProperty = false
+    
+    for (prop1, prop2) in zip(lhs.children, rhs.children) {
```
kyleve, line 64: Also check: If the `children` are the same length, return `.notEqual` if not.

ListableUI/Sources/Item/ItemContent.swift
```diff
@@ -545,13 +475,13 @@ public extension ItemContent {
...
 
 /// Provides a default implementation of `identifierValue` when self conforms to Swift's `Identifiable` protocol.
-@available(iOS 13.0, *)
-public extension ItemContent where Self:Identifiable
-{    
-    var identifierValue : ID {
-        self.id
-    }
-}
+//@available(iOS 13.0, *)
```
kyleve, line 478: Revert

ListableUI/Tests/ComparePropertiesTests.swift
```diff
@@ -0,0 +1,369 @@
...
+//  ListableUI-Unit-Tests
+//
+//  Created by Kyle Van Essen on 7/28/22.
+//
+
+@_spi(ListableInternal) import ListableUI
+import XCTest
+
+
+class ComparePropertiesTests : XCTestCase {
```
kyleve, line 12: Add some tests for arrays / dictionaries too.

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -0,0 +1,184 @@
...
+
+
+// MARK: Item / ItemContent Extensions
+
+extension Element {
+    
+    /// Converts the given `Element` into a Listable `Item` with the provided ID. You can use this ID
+    /// to scroll to or later access the item through the regular list access APIs.
+    ///
+    /// You many also optionally configure the item, setting its values such as the `onDisplay` callbacks, etc.
```
nononoah, line 20:
```suggestion
    /// You may also optionally configure the item, setting its values such as the `onDisplay` callbacks, etc.
```

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -0,0 +1,184 @@
...
+    ///         item.insertAndRemoveAnimations = .scaleUp
+    ///     }
+    ///     .background {
+    ///         Box(backgroundColor: ...).inset(...)
+    ///     }
+    ///     .selectedBackground(.tappable) {
+    ///         Box(backgroundColor: ...).inset(...)
+    ///     }
+    /// ```
+    public func listItem(
```
nononoah, line 36: I think this is the right direction, and only question if we should require the presence of an id.

ListableUI/Sources/IsEquivalent/EquivalentComparable.swift
```diff
@@ -0,0 +1,155 @@
...
+//
+
+import Foundation
+
+
+/// Used by the list to determine when the content of content has changed; in order to
+/// remeasure the content and re-layout the list.
+///
+/// ## Note
+/// You should rarely need to implement ``EquivalentComparable/isEquivalent(to:)-15tcq``
```
nononoah, line 15: `-15tcq`? Is this a gen thing?

ListableUI/Sources/IsEquivalent/EquivalentComparable.swift
```diff
@@ -0,0 +1,155 @@
...
+//  EquivalentComparable.swift
+//  ListableUI
+//
+//  Created by Kyle Van Essen on 11/28/21.
+//
+
+import Foundation
+
+
+/// Used by the list to determine when the content of content has changed; in order to
```
nononoah, line 11:
```suggestion
/// Used by the list to determine when the content of a member has changed; in order to
```

BlueprintUILists/Sources/Element+Item.swift
nononoah: I subsequently understood that this was meant to improve the builder experience. I have mixed feelings about that but do think it matches the standard folks have come to expect from SwiftUI.

nononoah (approved):

nononoah:
I agree with the core aspects of the direction here - the builders, the reversal of construction (from `Item` containing `Element` to `Element` modified to `Item`), and even some of the more controversial sugar that you've included. Thanks for backing this up with so much testing.

BlueprintUILists/Sources/Element+HeaderFooter.swift
```diff
@@ -0,0 +1,160 @@
...
+        var copy = self
+        copy.onTap = onTap
+        copy.content._backgroundProvider = background
+        return copy
+    }
+}
+
+
+
+/// Ensures that the `Equatable` initializer for `WrappedHeaderFooterContent` is called.
```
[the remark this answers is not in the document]
kyleve: This is needed for the result builders in particular – so it's not that we can provide an overload that's useful – it's that we need to signal to the compiler that this is the method set, etc to use for these passing types

BlueprintUILists/Sources/Element+Item.swift
kyleve: My 2c / thoughts are that, since you only really need to provide an ID as an optimization (eg if your list is changing a lot, or you're reading data back out – eg for reordering), the framework can already make a pretty good guess what you meant, so requiring it isn't super useful

ListableUI/Sources/IsEquivalent/EquivalentComparable.swift
kyleve: It's a DocC thing for disambiguating / referring to methods – I don't know what the exact bits mean, but you can repro it fairly commonly when writing DocC comments and relying on Xcode to autocomplete the references

BlueprintUILists/Sources/Element+Item.swift
kyleve: (And indeed, people get the ID wrong pretty often – eg just passing a `UUID()` every time – which is very destabilizing / a perf issue)

BlueprintUILists/Sources/Element+HeaderFooter.swift
```diff
@@ -0,0 +1,160 @@
...
+    /// configure the header / footer, setting its values such as the `onTap` callbacks, etc.
+    ///
+    /// You may also provide a background and pressed background as well as tap actions.
+    ///
+    /// ```swift
+    /// MyElement(...)
+    ///     .listHeaderFooter { header in
+    ///         header.onTap = { ... }
+    ///     }
+    ///     .background {
```
kyleve, line 28: This should be _before_ the `listHeaderFooter` bit.

BlueprintUILists/Sources/Element+HeaderFooter.swift
```diff
@@ -0,0 +1,160 @@
...
+    }
+    
+    /// TODO
+    public func onTap(
+        _ onTap : @escaping () -> (),
+        show background : @escaping () -> Element
+    ) -> Self {
+        var copy = self
+        copy.onTap = onTap
+        copy.content._backgroundProvider = background
```
kyleve, line 64: This should be the pressed background

BlueprintUILists/Sources/Element+HeaderFooter.swift
```diff
@@ -0,0 +1,160 @@
...
+    
+    public func isEquivalent(to other: Self) -> Bool {
+        isEquivalent(self, other)
+    }
+    
+    public var elementRepresentation: Element {
+        represented
+    }
+    
+    public var _backgroundProvider : () -> Element? = { nil }
```
kyleve, line 137: Consider SPIing these

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -0,0 +1,184 @@
...
+        _backgroundProvider(info)
+    }
+    
+    public var _selectedBackgroundProvider: (ApplyItemContentInfo) -> Element? = { _ in nil }
+    
+    public func selectedBackgroundElement(with info: ApplyItemContentInfo) -> Element? {
+        _selectedBackgroundProvider(info)
+    }
+    
+    public var reappliesToVisibleView: ReappliesToVisibleView {
```
kyleve, line 173: Consider: Make this an option too, with this being the default

ListableUI/Sources/IsEquivalent/CompareProperties.swift
```diff
@@ -0,0 +1,446 @@
...
+        return true
+    }
+}
+
+
+infix operator ||=
+
+fileprivate extension Bool {
+    
+    static func ||= (lhs : inout Bool, rhs : Bool) {
```
JaviSoto, line 443: Normally you'd want the right side of this to be lazily evaluated (`@autoclosure`). It won't matter the way it's being used here though.

ListableUI/Sources/IsEquivalent/CompareProperties.swift
```diff
@@ -0,0 +1,446 @@
...
+                
+            case .notEqual:
+                hadEquatableProperty ||= true
+                return .notEqual
+                
+            case .hasNoFields:
+                hadEquatableProperty ||= false
+                
+            case .error:
+                hadEquatableProperty ||= false
```
JaviSoto, line 134: Isn't this a no-op?

ListableUI/Sources/IsEquivalent/CompareProperties.swift
```diff
@@ -0,0 +1,446 @@
...
+            /// Othewise, we will recursively check its child values.
+            
+            let result = compareEquatableProperties(prop1.value, prop2.value)
+            
+            switch result {
+            case .equal:
+                hadEquatableProperty ||= true
+                
+            case .notEqual:
+                hadEquatableProperty ||= true
```
JaviSoto, line 127: Isn't this the same as setting it to true?

JaviSoto (commented):
Don't mind me, I was browsing Mastodon at 2am because I couldn't sleep and saw this PR and noticed a couple funny things 😄

ListableUI/Sources/IsEquivalent/CompareProperties.swift
kyleve: Ahh yeah, good idea.

ListableUI/Sources/IsEquivalent/CompareProperties.swift
kyleve: It is, but I figured for consistency of the method I'd just do it... I can convert these to `break`s as well.

ListableUI/Sources/IsEquivalent/CompareProperties.swift
kyleve: It is, yeah; I just sort of liked how the method looked by doing the same thing everywhere.

BlueprintUILists/Sources/Element+HeaderFooter.swift
kyleve: Wait no it shouldn't

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -585,7 +585,7 @@
 				GCC_C_LANGUAGE_STANDARD = gnu11;
 				GCC_DYNAMIC_NO_PIC = NO;
 				GCC_NO_COMMON_BLOCKS = YES;
-				GCC_OPTIMIZATION_LEVEL = 0;
+				GCC_OPTIMIZATION_LEVEL = s;
```
kyleve, line 588: - [ ] Revert this before merging

ListableUI/Sources/IsEquivalent/IsEquivalentContent.swift
kyleve: So uhh hi, its a year later; I think I still actually prefer just going with `IsEquivalent`. It's more to the point for the single constraint of the protocol, and it makes refinements make more sense. Eg, `IsKeyPathEquivalent` for a variant based on provided key paths...

kyleve: Ok going with `LayoutEquivalent`

BlueprintUILists/Sources/Element+Item.swift
kyleve: Fixed!

BlueprintUILists/Sources/Element+Item.swift
kyleve: Actually, lets just rip this out

CHANGELOG.md
```diff
@@ -6,10 +6,33 @@
 
 ### Added
```
kyleve, line 7: - [ ] Keypath equivalency

CHANGELOG.md
```diff
@@ -6,10 +6,33 @@
...
+
 ### Removed
 
 ### Changed
 
+- Definition of `isEquivalent(to:)` has been moved to `LayoutEquivalent`.
+
+- The `ListableBuilder` result builder is now `ListableArrayBuilder`.
+
+- In many cases, you no longer need to implement `isEquivalent(to:)`. Listable will traverse the `Equatable` properties on your contents to determine when content should be re-measured. To improve performance for long lists or complex content objects, you're still encouraged to either make those content objects conform to `Equatable`, or implement `isEquivalent(to:)`.
```
kyleve, line 34: - [ ] Remove

ListableUI/Sources/ListableBuilder.swift
```diff
@@ -86,3 +86,32 @@
...
+/// }
+/// ```
+///
+/// ## Links & Videos
+/// https://github.com/apple/swift-evolution/blob/main/proposals/0289-result-builders.md
+/// https://developer.apple.com/videos/play/wwdc2021/10253/
+/// https://www.swiftbysundell.com/articles/deep-dive-into-swift-function-builders/
+/// https://www.avanderlee.com/swift/result-builders/
+///
+@resultBuilder public enum ListableValueBuilder<ContentType> {
```
kyleve, line 108: `ListableOptionalValueBuilder`?

BlueprintUILists/Sources/ListableBuilder+Element.swift
```diff
@@ -0,0 +1,88 @@
...
+///     TestContent1() // An ItemContent
+///
+///     Element1() // An Element
+///     Element2() // An Element
+/// }
+/// ```
+public extension ListableArrayBuilder where ContentType == AnyItemConvertible {
+
+    /// Ensures that a well-formed error is presented when a non-Equatable or non-LayoutEquivalent element is provided.
+    @available(*, unavailable, message: "To be directly added to a List, an Element must conform to Equatable or LayoutEquivalent.")
```
kyleve, line 25: - [ ] TODO: Test this is indeed causing an error before merging.

BlueprintUILists/Sources/ListableBuilder+Element.swift
```diff
@@ -0,0 +1,88 @@
...
+    static func buildExpression<ElementType:Element>(_ element: ElementType) -> Component where ElementType:ListElementNonConvertible {
+        element.listElementNonConvertibleFatal()
+    }
+}
+
+
+public extension ListableOptionalBuilder where ContentType == AnyHeaderFooterConvertible {
+
+    /// Ensures that a well-formed error is presented when a non-Equatable or non-LayoutEquivalent element is provided.
+    @available(*, unavailable, message: "To be directly added to a List, an Element must conform to Equatable or LayoutEquivalent.")
```
kyleve, line 51: - [ ] TODO: Test this is indeed causing an error before merging.

ListableUI/Sources/IsEquivalent/CompareProperties.swift
kyleve: I removed the auto-mirror stuff, so this is gone now.

ListableUI/Sources/IsEquivalent/CompareProperties.swift
kyleve: I removed the auto-mirror stuff, so this is gone now.

BlueprintUILists/Sources/Element+Item.swift
kyleve: Hey its been over a year, I am going back on this and requiring the ID! If we see it really getting abused _again_, we can remove this.

kyleve: Actually, hmm, this _does_ make quite a few places less ergonomic. I'm going to ruminate on this one a bit more.

kyleve:
_Finally_ getting around to landing this!

BlueprintUILists/Sources/Element+Item.swift
kyleve: Ok, I went back on this again. Not doing an ID.

PR #416, opened 2022-07-27
[WIP DNR] Minimum Requested Spacing

kyleve:

### Checklist


ListableUI/Sources/Layout/Table/TableListLayout.swift
```diff
@@ -429,6 +429,8 @@ final class TableListLayout : ListLayout
         viewWidth : CGFloat,
```
kyleve, line 429: TODO: We're not always setting `lastAddedSpacing` when we set `contentBottom`, and we should

BlueprintUILists/Sources/Element+Item.swift
```diff
@@ -81,6 +83,34 @@ extension Element where Self:IsEquivalentContent {
 }
 
 
+extension Element where Self:BlueprintItemContent {
```
kyleve, line 86: This can go away its old

PR #228, opened 2020-10-28
Fix autolayout sizing for items and headers/footers

kyleve:
(still need tests)

PR #418, opened 2022-07-31
Dismiss swipe actions when touching outside the active cell

kyleve:
I realized this behavior was missing: When touching down outside of the actively swiped cell, we should dismiss the active swipe, and let the touch continue to scroll the scroll view.

### Checklist


kylebshr:
Huh I definitely added this before, it must have gotten removed? https://github.com/kyleve/Listable/pull/130/files#diff-7745c8a7f6e1a90d892771b03ba9a41d35e6ad17de57ad094ee044eb08dc45a4R39

kylebshr:
Removed here: https://github.com/kyleve/Listable/pull/228/files#diff-501c1250e3f754e63a56d3119275705080d224bd3197c74d6f0b2dd70e1f7657L36

In favor of this: https://github.com/kyleve/Listable/pull/228/files#diff-e63bc75982fd7d7cfe6fafc5ac2ddd1507af287f76c4d05716a306c5cbad550dR225

Which is still on main: https://github.com/kyleve/Listable/blob/0f796bb571543b51262bcacf7eff9bfaf5685884/ListableUI/Sources/ListView/ListView.Delegate.swift#L284

But does it not work?

kyleve:
That only works when dragging begins, vs on touch down with no drag - which is what this implements. Eg see this video of Mail.app, each time the swipe dismisses, I'm only touching down, not dragging at all.

https://user-images.githubusercontent.com/327847/182009411-428826fd-4781-4833-851f-50ad8a246b44.MOV

kylebshr:
Got itttt makes sense!

kyleve:
@kylebshr if you have a chance, mind giving this a review?

ListableUI/Sources/Internal/ItemCell.swift
```diff
@@ -179,6 +185,24 @@ final class LiveCells {
...
+            }
+        }
+        
+        return cell?.cell
+    }
+    
+    var activeSwipeCell : AnyItemCell? {
+        first { cell in
+            cell.areSwipeActionsVisible
+        }
```
n8chur, line 203: nit:
```suggestion
        first(where: \.areSwipeActionsVisible)
```

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -80,7 +82,14 @@ public final class ListView : UIView, KeyboardObserverDelegate
         self.delegate.layoutManager = self.layoutManager
         
         self.keyboardObserver.add(delegate: self)
-                
+        
+        self.closeActiveSwipesGesture.addTarget(self, action: #selector(closeActiveSwipeGestureIfNeeded))
+        self.addGestureRecognizer(closeActiveSwipesGesture)
```
n8chur, line 87: Can swipe action view use gesture recognizers to observe touch? If they do, is it possible to activate them with this approach or does the `closeActiveSwipesGesture` capture the touch down preventing interaction?

It seems like we may need to ensure that a tap down inside the swipe action space doesn't prevent any gestures from occurring, right?

ListableUI/Sources/Internal/ItemCell.swift
kyleve: TIL, I thought this only existed for map!

ListableUI/Sources/ListView/ListView.swift
kyleve: That's what the `shouldRecognize` does – it'll only occur if there is (a) an active swipe, and (b) the touch is not inside the active swipe view

ListableUI/Sources/Internal/ItemCell.swift
n8chur: Works with any closure!

kyleve: Magic

ListableUI/Sources/ListView/ListView.swift
n8chur: Ah, I missed (b) down below.

It looks like that check is detecting whether the tap is in the cell—should i be checking that it's in the swipe action space? That way tapping on the other cell content would dismiss it?

n8chur (approved):

ListableUI/Sources/ListView/ListView.swift
kyleve: Ah yeah, I think I can do that, and it looks to be how UIKit works as well

kyleve: Done

PR #421, opened 2022-08-03
Add .toggles selection type for toggle-style selections

kyleve:
This allows the toggle-style interaction; where tapping once selects the row, and tapping again deselects it.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.Delegate.swift
+++ b/ListableUI/Sources/ListView/ListView.Delegate.swift
@@ -52,7 +52,22 @@ extension ListView
             
             let item = self.presentationState.item(at: indexPath)
             
-            return item.anyModel.selectionStyle.isSelectable
+            if case .toggles = item.anyModel.selectionStyle {
+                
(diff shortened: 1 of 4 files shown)


ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -52,7 +52,22 @@ extension ListView
...
             
-            return item.anyModel.selectionStyle.isSelectable
+            if case .toggles = item.anyModel.selectionStyle {
+                
+                if item.isSelected {
+                    item.set(isSelected: false, performCallbacks: true)
+                    collectionView.deselectItem(at: indexPath, animated: false)
+                    item.applyToVisibleCell(with: self.view.environment)
+                    
+                    self.performOnSelectChanged()
```
tristanbiesecker, line 62: Would this result in `Item.onDeselect` being called?

tristanbiesecker (approved):
Thanks for doing this so quickly!

PR #424, opened 2022-08-17
Add Behavior.decelerationRate and ListPagingBehavior.firstVisibleItem…

millerjasonw:
Adding Behavior.decelerationRate for snappier paging and ListPagingBehavior.firstVisibleItemCentered to center snapped items.

### Checklist


ListableUI/Sources/Layout/ListLayout/ListLayout.swift
```diff
@@ -378,10 +378,11 @@ extension AnyListLayout
 {
     func onDidEndDraggingTargetContentOffset(
         for targetContentOffset : CGPoint,
-        velocity : CGPoint
+        velocity : CGPoint,
+        visibleContentFrame: CGRect = .zero
```
millerjasonw, line 382: Any alternative already available to ListLayout for knowing the visible bounds? Haven't found any, including through content.

kyleve, line 382: Should this have a default? Imo, likely not – since that's sort of meaningless, and likely wrong.

kyleve: I can't think of anything else! But I do wonder if you need to pass in the whole rect, or just the size?

ListableUI/Sources/LayoutDirection.swift
```diff
@@ -249,3 +249,74 @@ extension LayoutDirection
...
     }
 }
+
+
+// MARK: KeyPaths
+
+extension LayoutDirection
+{
+
+    var point: WritableKeyPath<CGPoint, CGFloat>
```
kyleve, line 259: It's a bit odd to me a single value (the x or y of a point) is called point – I'd expect this keyPath to point to, well, a `CGPoint`.

ListableUI/Sources/LayoutDirection.swift
```diff
@@ -249,3 +249,74 @@ extension LayoutDirection
...
+    {
+        switch self {
+        case .horizontal: return \.x
+        case .vertical: return \.y
+        }
+    }
+    
+    var minPoint: KeyPath<CGRect, CGFloat>
```
kyleve, line 267: Clever, I love these!

ListableUI/Sources/LayoutDirection.swift
kyleve: (I do wonder if we could just call these as `{foo}Y` (or `{foo}X in the case that `vertical` returns an `x), since that's how the other APIs work: Assume you're working in "default" land, where the default is `vertical`.)

kyleve (commented):
Looks really good, thanks for doing this! A few comments / questions, mostly on naming.

Also:
- [ ] Update `CHANGELOG.md`

ListableUI/Sources/Layout/ListLayout/ListLayout.swift
```diff
@@ -390,15 +391,22 @@ extension AnyListLayout
...
+
+        var newTargetContentOffset = targetContentOffset
+        switch self.pagingBehavior {
+        case .firstVisibleItemEdge:
+            let itemOffset = item.defaultFrame[keyPath: direction.minPoint]
+            newTargetContentOffset[keyPath: direction.point] = itemOffset - padding[keyPath: direction.openingInset]
+            newTargetContentOffset[keyPath: direction.otherAxisPoint] = .zero
+        case .firstVisibleItemCentered:
+            let itemOffset = item.defaultFrame[keyPath: direction.midPoint]
+            newTargetContentOffset[keyPath: direction.point] = itemOffset - visibleContentFrame.size[keyPath: direction.size] / 2
```
kyleve, line 404: You likely want to round this to full value in case the `/ 2` isn't on a screen pixel.

ListableUI/Sources/Layout/ListLayout/ListLayout.swift
millerjasonw: 👍 size works for me. Will do.

millerjasonw: Yeah, I wasn't fully committed to even having the arg, so put in temporarily. Will do.

ListableUI/Sources/LayoutDirection.swift
millerjasonw: Maybe "position"? I added these to get away from a situation where X returns Y, height returns width, etc. It's tripped me up once already.

ListableUI/Sources/Behavior.swift
```diff
@@ -88,7 +93,26 @@ extension Behavior
...
+    {
+        public static var normal: DecelerationRate {
+            .init(rawValue: UIScrollView.DecelerationRate.normal.rawValue)
+        }
+
+        public static var fast: DecelerationRate {
+            .init(rawValue: UIScrollView.DecelerationRate.fast.rawValue)
+        }
+
+        public var rawValue: CGFloat
```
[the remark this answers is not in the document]
millerjasonw: Yeah wondered about that – it'd commit us to having to support a float even when the underlying component changes/no longer does.

millerjasonw:
How're we looking now @kyleve ?

kyleve:
> How're we looking now @kyleve ?

Oops sorry, looking!

ListableUI/Sources/LayoutDirection.swift
```diff
@@ -249,3 +249,74 @@ extension LayoutDirection
...
     }
 }
+
+
+// MARK: KeyPaths
+
+extension LayoutDirection
+{
+
+    var coordinate: WritableKeyPath<CGPoint, CGFloat>
```
kyleve, line 259: Sorry for sticking here, but I would prefer if we leave these as referring to the "vertical" layout vs coming up with a new naming scheme, since the existing methods in here all refer to the names as their "vertical" equivalents, eg:

```
    public func y(for point : CGPoint) -> CGFloat
    {
        switch self {
        case .vertical: return point.y
        case .horizontal: return point.x
        }
    }
```

So I would just call this `pointX`, etc.

kyleve: (There's no great naming scheme here because both referring to them as their vertical names can be confusing, but so can the Blueprint model, for example, where we refer to things as axis and cross-axis – but I'd rather we remain consistent with the existing names)

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -251,7 +251,14 @@ public final class ListView : UIView, KeyboardObserverDelegate
         
         self.collectionView.canCancelContentTouches = self.behavior.canCancelContentTouches
         self.collectionView.delaysContentTouches = self.behavior.delaysContentTouches
-        
+
+        self.collectionView.decelerationRate = {
```
kyleve, line 255: Any idea if setting this can mess with an in-progress scroll event? `UIScrollView` has weird behaviors where setting a property, even to the same value, can mess with scrolling. Eg this is why Blueprint does [this](https://github.com/square/Blueprint/blob/main/BlueprintUICommonControls/Sources/ScrollView.swift#L320-L350) when configuring its scroll view, and we do the [same](https://github.com/kyleve/Listable/blob/7c071bacb1899be18428c62b9aaef40c0bf3db76/ListableUI/Sources/ListView/ListView.swift#L317-L323) when setting content inset.

ListableUI/Tests/Layout/ListLayout/ListLayoutTests.swift
```diff
@@ -25,7 +25,8 @@ final class AnyListLayoutTests : XCTestCase {
             XCTAssertEqual(
                 layout.onDidEndDraggingTargetContentOffset(
                     for: CGPoint(x: 0, y: 0),
-                       velocity: CGPoint(x: 0, y: 1)
+                    velocity: CGPoint(x: 0, y: 1),
+                    visibleContentSize: .zero
```
kyleve, line 29: 1) Is passing `.zero` valid/correct here? Eg I'd expect this value in production to always be non-zero.

2) Can you add some tests for your new paging behavior? Looks like the existing tests are only testing `.none` and `.firstVisibleItemEdge`.

ListableUI/Sources/LayoutDirection.swift
millerjasonw: Moot. Pulled out the keypaths.

ListableUI/Sources/LayoutDirection.swift
millerjasonw: Done.

ListableUI/Sources/ListView/ListView.swift
millerjasonw: Done. Happy to play it safe.

ListableUI/Tests/Layout/ListLayout/ListLayoutTests.swift
millerjasonw: Done.

millerjasonw:
How're we looking now @kyleve ?

ListableUI/Sources/Layout/ListLayout/ListLayout.swift
millerjasonw: Done.

PR #434, opened 2022-10-15
Fix Update Queueing

kyleve:
This fixes an issue where, it seems like, in some cases, `performBatchUpdates(_:completion:)` will call its update block non-synchronously – which means if multiple updates end up getting submitted in quick succession; the latter update can end up in an invalid state and crash. The fix is to ensure we queue critical update work in our updates queue. The primary changes here are in `ListChangesQueue` and `ListView`.

### Checklist

### DIFF
--- a/Demo/Sources/Demos/Demo Screens/UpdateFuzzingViewController.swift
+++ b/Demo/Sources/Demos/Demo Screens/UpdateFuzzingViewController.swift
@@ -1,0 +1,172 @@
+//
+//  UpdateFuzzingViewController.swift
+//  Demo
+//
+//  Created by Kyle Van Essen on 10/17/22.
+//  Copyright © 2022 Kyle Van Essen. All rights reserved.
+//
+
+import UIKit
+import ListableUI
+import BlueprintUILists
+import BlueprintUI
+import BlueprintUICommonControls
+import EnglishDictionary
+
+
+final class UpdateFuzzingViewController : ListViewController {
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        self.navigationItem.rightBarButtonItems = [
+            UIBarButtonItem(title: "Start", style: .plain, target: self, action: #selector(startFuzzing)),
+            UIBarButtonItem(title: "Stop", style: .plain, target: self, action: #selector(stopFuzzing)),
+        ]
+    }
+    
+    private lazy var dictionary = EnglishDictionary.dictionary
+    
+    var isFilterOn : Bool = false
+    
+    override func configure(list: inout ListProperties) {
+        
+        list.layout = .table {
+            
+            $0.bounds = .init(
+                padding: UIEdgeInsets(top: 0.0, left: 20.0, bottom: 20.0, right: 20.0),
+                width: .atMost(600.0)
+            )
+            
+            $0.layout.set {
+                $0.sectionHeaderBottomSpacing = 10.0
+                $0.itemSpacing = 7.0
+                $0.interSectionSpacingWithNoFooter = 10.0
+            }
+        }
+        
+        list += self.dictionary.wordsByLetter.skipHalfMap(skip: isFilterOn) { letter in
+            Section(letter.letter) {
+                letter.words.skipHalfMap(skip: isFilterOn) { word in
+                    Item(
+                        WordRow(title: word.word, detail: word.description),
+                        sizing: .thatFits(.init(.atMost(250.0)))
+                    )
+                }
+            } header: {
+                SectionHeader(title: letter.letter)
+            }
+        }
+    }
+    
+    private var timer : Timer? = nil
+    
+    @objc private func stopFuzzing() {
+        timer?.invalidate()
+        timer = nil
+    }
+    
+    @objc private func startFuzzing() {
+        guard timer == nil else { return }
+        
+        timer = Timer(timeInterval: 0.05, repeats: true) { [weak self] _ in
+            guard let self = self else { return }
+            
+            self.isFilterOn.toggle()
+            
+            print("Reloading content...")
+            
+            self.reload(animated: true)
+        }
+        
+        RunLoop.current.add(timer!, forMode: .tracking)
+    }
+}
+
+
+fileprivate extension Array {
+    
+    func skipHalfMap<Mapped>(skip : Bool, using block : (Element) -> Mapped) -> [Mapped] {
+        
+        indexedCompactMap { index, element in
+            if index % 2 == 0 && skip {
+                return nil
+            } else {
+                return block(element)
+            }
+        }
+        
+    }
+    
+    func indexedCompactMap<Mapped>(_ map: (Int, Element) -> Mapped?) -> [Mapped] {
+
+        var mapped = [Mapped]()
+
+        for index in indices {
+            if let value = map(index, self[index]) {
+                mapped.append(value)
+            }
+        }
+
+        return mapped
+    }
+}
+
+
+fileprivate struct SectionHeader : BlueprintHeaderFooterContent, Equatable
+{
+    var title : String
+    
+    // MARK: BlueprintItemElement
+    
+    var elementRepresentation: Element {
+        return Box(
+            backgroundColor: UIColor(white: 0.85, alpha: 1.0),
+            cornerStyle: .rounded(radius: 10.0),
+            wrapping: Inset(
+                top: 10.0,
+                bottom: 10.0,
+                left: 20.0,
+                right: 20.0,
+                wrapping: Label(text: self.title) {
+                    $0.font = .systemFont(ofSize: 32.0, weight: .bold)
+            })
+        )
+    }
+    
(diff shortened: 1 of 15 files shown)


ListableUI/Sources/Internal/Diff/SectionedDiff.swift
```diff
@@ -404,7 +404,38 @@ extension SectionedDiff.ItemChanges.Updated : Equatable where Item:Equatable {}
 extension SectionedDiff.ItemChanges.NoChange : Equatable where Item:Equatable {}
 
 
-extension SectionedDiff.SectionChanges
+extension SectionedDiff {
+    
+    func transform<Mapped>(
```
kyleve, line 409: This moved around so that I can add the validation added below, so we fail in a predictable way for this failure mode, instead of harder to debug out of range issues.

ListableUI/Sources/ListView/ListChangesQueue.swift
```diff
@@ -13,11 +13,54 @@ import Foundation
 final class ListChangesQueue {
```
kyleve, line 13: TODO: Check for memory leaks
- [ ] Sync
- [ ] Async
- [ ] Removing All

ListableUI/Sources/Internal/Diff/SectionedDiff.swift
```diff
@@ -404,33 +404,66 @@ extension SectionedDiff.ItemChanges.Updated : Equatable where Item:Equatable {}
 extension SectionedDiff.ItemChanges.NoChange : Equatable where Item:Equatable {}
 
 
-extension SectionedDiff.SectionChanges
+extension SectionedDiff {
```
kyleve, line 407: This moved so that we have access to `self.old`; to compare the counts of then old and new diffable sections.

ListableUI/Sources/Internal/Diff/SectionedDiff.swift
```diff
@@ -404,33 +404,66 @@ extension SectionedDiff.ItemChanges.Updated : Equatable where Item:Equatable {}
...
+    
+    /// Takes the content of the `input` array, and transforms it using the diff's changes
+    /// into a newly returned array, creating, moving, or updating the content as required.
+    func transform<Mapped>(
+        input : [Mapped],
+        removed : (Section, Mapped) -> (),
+        added : (Section) -> Mapped,
+        moved : (Section, Section, ItemChanges, inout Mapped) -> (),
+        noChange : (Section, Section, ItemChanges, inout Mapped) -> (),
+        mappedItemCount : (Mapped) -> Int,
```
kyleve, line 417: These two parameters were added for the precondition.

Demo/Sources/Demos/Demo Screens/UpdateFuzzingViewController.swift
```diff
@@ -0,0 +1,172 @@
...
+
+import UIKit
+import ListableUI
+import BlueprintUILists
+import BlueprintUI
+import BlueprintUICommonControls
+import EnglishDictionary
+
+
+final class UpdateFuzzingViewController : ListViewController {
```
kyleve, line 17: This is largely a copy of `SearchableDictionaryViewController`, which I wanted because there's a lot of data in there, so it simulates a "real" long list.

ListableUI/Tests/Internal/Diff/ArrayDiffTests.swift
```diff
@@ -359,7 +359,7 @@ class ArrayDiffTests: XCTestCase
     
     func test_transform_with_random_mutations()
     {
-        let iterations : Int = 500
+        let iterations : Int = 2000
```
kyleve, line 362: Intentional, cover more cases.

ListableUI/Tests/XCTestCaseAdditions.swift
```diff
@@ -10,6 +10,53 @@ import XCTest
...
 {
+    ///
+    /// Call this method to show a view controller in the test host application
+    /// during a unit test. The view controller will be the size of host application's device.
+    ///
+    /// After the test runs, the view controller will be removed from the view hierarchy.
+    ///
+    /// A test failure will occur if the host application does not exist, or does not have a root view controller.
+    ///
+    public func show<ViewController: UIViewController>(
```
kyleve, line 21: Copied from Market and Workflow where it already lives

ListableUI/Sources/ListView/ListChangesQueue.swift
```diff
@@ -8,13 +8,99 @@
...
-/// Used to queue updates into a list view.
-/// Note: This type is only safe to use from the main thread.
+/// A queue used to synchronized and serialize changes made to the backing collection view,
+/// to work around either bugs or confusing behavior.
+///
+/// ## Handling Re-ordering (`isQueuingForReorderEvent`)
+/// Collection View has an issue wherein if you perform a re-order event, and then within
+/// the same runloop, deliver an update to the collection view as a result of that re-order event
+/// that removes a row or section, the collection view will crash because it's internal index path
+/// cache / data model has  not yet been updated. Thus, in `collectionView(_:,moveItemAt:,to:)`,
```
n8chur, line 18:
```suggestion
/// cache / data model has not yet been updated. Thus, in `collectionView(_:,moveItemAt:,to:)`,
```

ListableUI/Sources/ListView/ListChangesQueue.swift
```diff
@@ -32,68 +118,173 @@ final class ListChangesQueue {
...
 }
 
 
 extension ListChangesQueue {
-    
+        
     final class Operation {
         
-        fileprivate(set) var state : State
+        fileprivate(set) var kind : Kind
```
n8chur, line 199: Is this the only property? could we flatten this so `Operation` is the `enum`?

ListableUI/Sources/ListView/ListChangesQueue.swift
```diff
@@ -32,68 +118,173 @@ final class ListChangesQueue {
...
+            "ListChangesQueue must run on main thread. Instead, it was \(Thread.current)."
+        )
+    }
 }
 
 
 extension ListChangesQueue {
-    
+        
     final class Operation {
```
n8chur, line 197: Could/should this be `fileprivate`?

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1009,80 +1009,105 @@ public final class ListView : UIView, KeyboardObserverDelegate
...
-                slice: visibleSlice,
-                reason: .wasUpdated,
                 animated: reason.animated,
-                dependencies: dependencies,
-                updateCallbacks: updateCallbacks,
-                loggable: self
+                updateBackingData: updateBackingData,
+                collectionViewUpdateCompletion: {
+                    completion.finished()
+                },
```
n8chur, line 1082: Nit:
```suggestion
                collectionViewUpdateCompletion: completion.finished,
```

ListableUI/Sources/ListView/ListChangesQueue.swift
```diff
@@ -32,68 +118,173 @@ final class ListChangesQueue {
...
         
-        let first = self[0]
-        
-        self.remove(at: 0)
+        private var isFinished : Bool = false
         
-        return first
+        /// Invoked by callers when their async work completed.
+        /// If this method is called more than once, a fatal error occurs.
+        func finished() {
```
n8chur, line 282: Nit:
```suggestion
        func finish() {
```

n8chur (approved):

ListableUI/Sources/ListView/ListChangesQueue.swift
kyleve: Hmm, I don't think we can (in a super ergonomic way), because we need the outer reference `class` type to mutate as states transition – otherwise we have to deal with mutating the array of queued events, which is a bit more messy, IMO.

ListableUI/Sources/ListView/ListChangesQueue.swift
kyleve: I think I can do that if I introduce a `count` to the queue, so we can also hide the `waiting` array – let me check

ListableUI/Sources/ListView/ListChangesQueue.swift
kyleve: Good call

ListableUI/Sources/ListView/ListChangesQueue.swift
kyleve: Committing locally

ListableUI/Sources/ListView/ListChangesQueue.swift
kyleve: Yep, this worked!

ListableUI/Sources/ListView/ListView.swift
kyleve: Good idea, fixing locally

PR #443, opened 2022-12-19
Extension Safety: Keyboard observing

lickel:
This makes `KeyboardObserver` a public class. 
There's also a public method on ListView which registers the keyboard observer.

CocoaPods is _effectively_ unchanged, as we are still using `+load`.
Adding this to SPM would be somewhat difficult as there's a circular dependency between the Swift and ObjC files.

### Checklist


ListableUI/Sources/KeyboardObserver/KeyboardObserver.swift
```diff
@@ -8,8 +8,8 @@
 import UIKit
 
 
-protocol KeyboardObserverDelegate : AnyObject {
-    
+public protocol KeyboardObserverDelegate : AnyObject {
```
kyleve, line 11: Is it worth hiding this behind `@_spi(ListableKeyboard)` or something perhaps?

kyleve (approved):

ListableUI/Sources/KeyboardObserver/KeyboardObserver.swift
lickel: As long as I can import it in MarketUI, that works for me

PR #448, opened 2023-01-12
[WIP DNR] Attempt to fix scrolling crash

kyleve:

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListChangesQueue.swift
+++ b/ListableUI/Sources/ListView/ListChangesQueue.swift
@@ -212,7 +212,13 @@ extension ListChangesQueue {
         
         /// Invoked by callers when their async work completed.
         /// If this method is called more than once, a fatal error occurs.
-        func finish(_ action : () -> () = {}) {
+        func finish() {
+            finish({})
+        }
+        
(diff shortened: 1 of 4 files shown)


ListableUI/Sources/ListView/ListChangesQueue.swift
```diff
@@ -212,9 +212,11 @@ extension ListChangesQueue {
         
         /// Invoked by callers when their async work completed.
         /// If this method is called more than once, a fatal error occurs.
-        func finish() {
+        func finish(_ action : () -> () = {}) {
```
gracekryu, line 215: Why do we need to add the action? Should we be concerned with action() being completed after `isFinished` is set to true?

kyleve: I'll likely remove this; I was thinking it might be nice to shape the `finish` API like:

```
operation.finish {
   /// Work the caller should do before finishing
}
```
To force ordering, but its probably not super useful given it's internal.

kyleve:
More speculative / lower-effort workaround here: https://github.com/square/Listable/pull/454

If this fixes the issue, I'll work on this further.

PR #454, opened 2023-02-14
Low effort attempt at fixing scrolling crash

kyleve:
Attempt to fix a crash that could occur when a scroll event was enqueued behind other major updates to the list view's content, which could result in index paths and section indexes becoming invalid.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -472,10 +472,20 @@ public final class ListView : UIView
         // Make sure the item identifier is valid.
 
         guard let toIndexPath = self.storage.allContent.firstIndexPathForItem(with: item) else {
+            completion(false)
             return false
         }
         
         return self.preparePresentationStateForScroll(to: toIndexPath) {
+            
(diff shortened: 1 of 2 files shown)


n8chur (approved):

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -472,10 +472,20 @@ public final class ListView : UIView
         // Make sure the item identifier is valid.
 
         guard let toIndexPath = self.storage.allContent.firstIndexPathForItem(with: item) else {
+            completion(false)
```
robmaceachern, line 475: What does the bool in `ScrollCompletion` represent?

kyleve: It's the standard UIView animation completion `Bool`, eg did this event complete or not. So it'll be false if (a) the animation is interrupted, or (b) the row/section is invalid.

kyleve: (No one ever actually like, reads that value for all intents and purposes, but that's what it's for)

robmaceachern (approved):

PR #456, opened 2023-02-23
weak self for a callback retained by a timer

zadr:
`ListView` is a `class` that has an internal timer that fires when sources change.

This timer has a callback that is triggered when the timer fires.

In the callback, since `ListView` is a class, we ultimately capture a reference `self`.

Since a `Timer` will keep a strong reference to its target (a `ReloadTimer` instance) and the target of the timer owns the block that captures `self` we wind up in a situation where a list can be kept alive a runloop longer than expected.

This is pretty harmless in apps, but I saw some non-determinism when looking at the memory graph for some KIF integration tests that I think ultimately comes back to this `Timer`.

### Checklist


kyleve (commented):
LGTM, nice catch

PR #461, opened 2023-03-18
Don't load background views unless they are requested

kyleve:
Makes view debugging easier and allocates fewer views when background views aren't needed!

### Checklist


ListableUI/Sources/Internal/HeaderFooterContentView.swift
```diff
@@ -21,8 +21,37 @@ final class HeaderFooterContentView<Content:HeaderFooterContent> : UIView
...
+        self.insertSubview(background, belowSubview: self.content)
+        
+        self.backgroundIfLoaded = background
+        
+        updateIsTappable()
+        
+        return background
+    }()
+    
+    private(set) var backgroundIfLoaded : Content.BackgroundView?
```
robmaceachern, line 38: I wonder if there's a way to spell this with a property wrapper that would avoid duplicating every property.

PR #462, opened 2023-03-29
Update swipe action view style when style changes

robmaceachern:
- Fixed an issue that could cause the wrong swipe action view style being applied when multiple styles were used in a single list.

Discovered in the demo:
image

### Checklist

### DIFF
--- a/ListableUI/Sources/Internal/DefaultSwipeView.swift
+++ b/ListableUI/Sources/Internal/DefaultSwipeView.swift
@@ -51,7 +51,14 @@ public final class DefaultSwipeActionsView: UIView, ItemContentSwipeActionsView 
 
     private var firstAction: SwipeAction?
     private var didPerformAction: SwipeAction.CompletionHandler
-    private let style: Style
+    
+    private var style: Style {
+        didSet {
+            if style != oldValue {
+                setNeedsLayout()
+            }
+        }
+    }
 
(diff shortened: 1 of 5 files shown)


ListableUI/Sources/Internal/DefaultSwipeView.swift
```diff
@@ -134,8 +141,12 @@ public final class DefaultSwipeActionsView: UIView, ItemContentSwipeActionsView
...
 
-    public func apply(actions: SwipeActionsConfiguration) {
-        if actionButtons.count != actions.actions.count {
+    public func apply(actions: SwipeActionsConfiguration, style: Style) {
+        let styleUpdateRequired = style != self.style
+        
+        self.style = style
+                
+        if actionButtons.count != actions.actions.count || styleUpdateRequired {
             actionButtons.forEach { $0.superview?.removeFromSuperview() }
```
robmaceachern, line 150: This will cause the button views to be blown away on a style update. We can be a bit more surgical about this and try to update the views instead but it didn't seem worth it. Can change though.

kyleve: Agree, not worth it. it shouldn't ever really change within the context of a single list.

kyleve (approved):

PR #465, opened 2023-04-07
Register callbacks for drag begin and deceleration end events

colealanroberts:
This PR introduces two public callbacks available on `ListStateObserver` for observing drag begin and drag end events.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.Delegate.swift
+++ b/ListableUI/Sources/ListView/ListView.Delegate.swift
@@ -298,11 +298,23 @@ extension ListView
             self.view.liveCells.perform {
                 $0.closeSwipeActions()
             }
+            
+            ListStateObserver.perform(self.view.stateObserver.onBeginDrag, "Will Begin Drag", with: self.view) { _ in
+                ListStateObserver.BeginDrag(
+                    positionInfo: self.view.scrollPositionInfo
+                )
+            }
         }
         
         func scrollViewDidEndDecelerating(_ scrollView: UIScrollView)
         {
             self.view.updatePresentationState(for: .didEndDecelerating)
+            
+            ListStateObserver.perform(self.view.stateObserver.onDidEndDeceleration, "Did End Deceleration", with: self.view) { _ in
+                ListStateObserver.DidEndDeceleration(
+                    positionInfo: self.view.scrollPositionInfo
+                )
+            }
         }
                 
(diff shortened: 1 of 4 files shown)


bchessin (approved):

ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -353,6 +360,13 @@ extension ListView
             withVelocity velocity: CGPoint,
             targetContentOffset: UnsafeMutablePointer<CGPoint>
         ) {
+            
+            ListStateObserver.perform(self.view.stateObserver.onDidEndDrag, "Did End Drag", with: self.view) { _ in
```
kyleve, line 364: I think we might want this in the `didEndDragging` callback, or `didEndDecelerating` – since `scrollViewWillEndDragging` is called when the user lifts their finger, not when the drag event fully completes.

colealanroberts: Moved to `didEndDragging`!

ListableUI/Sources/ListView/ListView.Delegate.swift
```diff
@@ -298,11 +298,23 @@ extension ListView
...
+                    positionInfo: self.view.scrollPositionInfo
+                )
+            }
         }
         
         func scrollViewDidEndDecelerating(_ scrollView: UIScrollView)
         {
             self.view.updatePresentationState(for: .didEndDecelerating)
+            
+            ListStateObserver.perform(self.view.stateObserver.onDidEndDeceleration, "Did End Deceleration", with: self.view) { _ in
```
kyleve, line 313: So question about this: I think for your purposes; you likely only want to call this / use this value to disable your auto-advance when the end of the deceleration is from a user-driven scroll event – eg _not_ when its from a programmatic event such as your auto-advance scroll. Should we be passing a flag to indicate that in `ListStateObserver.DidEndDeceleration`?

colealanroberts: @kyleve I've confirmed that `scrollViewDidEndDecelerating(_:)` isn't called when performing an auto-scroll action. It does however get called when interacting with the backing UIScrollView through user-initiated touch events.

kyleve (approved):
Thanks!

bchessin (approved):

PR #466, opened 2023-04-09
iOS 16.4 First Responder Bug

kyleve:
Works around a regression in iOS 16.4, but overriding the broken bits of `UICollectionView` with our own implementation. Please see `ListView+iOS16.4Workaround.swift` for a full discussion of the fix.

### Checklist


ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -0,0 +1,348 @@
...
+            /// methods (when converted to their final C representation) have two implicit arguments,
+            /// `self`, the object, and `SEL`, the selector.
+            typealias SuperFunction = @convention(c) (AnyObject, Selector) -> Bool
+            
+            /// Make the pointer we got back into a Swift-callable function.
+            let super_function = unsafeBitCast(super_impl, to: SuperFunction.self)
+            
+                   
+            /// This bug was fixed in iOS 16.5.
+            if #available(iOS 16.5, *) {
```
kyleve, line 263: - [ ] Confirm before merging

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -0,0 +1,348 @@
...
+        override init(
+            frame: CGRect,
+            collectionViewLayout layout: UICollectionViewLayout
+        ) {
+            super.init(frame: frame, collectionViewLayout: layout)
+        }
+        
+        required init?(coder: NSCoder) { fatalError() }
+        
+        @available(*, unavailable, message: "Please use performBatchUpdates:changes:completion:.")
```
n8chur, line 135:
```suggestion
        @available(*, unavailable, message: "Please use performBatch(_:changes:completion:).")
```

ListableUI/Sources/Internal/PresentationState/PresentationState.HeaderFooterState.swift
```diff
@@ -12,6 +12,12 @@ import UIKit
 protocol AnyPresentationHeaderFooterState : AnyObject
 {
     var anyModel : AnyHeaderFooter { get }
+    
+    var kind : SupplementaryKind { get }
+    
+    var oldIndexPath : IndexPath? { get }
```
n8chur, line 18: Can we document what this means and why it's useful?

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -0,0 +1,348 @@
...
+            /// `SupplementaryContainerView`, that's the only kind of supplementary view Listable has.
+            /// If for some other weird reason its not that, bail out.
+            guard let supplementaryView = owningReusableView as? SupplementaryContainerView else {
+                return super_function(self, selector)
+            }
+            
+            /// Ok so beyond this, we need to get it right; otherwise we'll hit this internal assert:
+            /// ```
+            /// 'The first responder contained inside of a deleted section or item refused to resign.
+            /// ```
```
n8chur, line 301: I'm having a hard time deciphering the hopper data, but I was expecting to potentially see references to `_firstResponderViewKind`, `finalIndexPathForInitialIndexPath`, and/or `_firstResponderViewType` in our implementation section if we're trying to mirror the internals of that implementation. I suppose it makes sense that we've deferred to super in any case that would normally need it, but how certain are we that the codepath that falls through here doesn't need to consider those properties? Are we pretty certain the logic below accounts for all necessary state here?

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -0,0 +1,348 @@
...
+        ///     rax = [rax retain];
+        ///     goto loc_36d6bf;
+        ///
+        /// loc_36d628:
+        ///     rbx = 0x0;
+        ///     goto loc_36d6d1;
+        /// }
+        ///
+        /// ```
+        @objc var _isFirstResponderInDeletedSectionOrItem : Bool {
```
n8chur, line 233: Do we know if this is ever called outside of a `performBatchUpdates` call?

It looks like if it were to get called out of `performBatchUpdates` we'd hit an assertion (only in DEBUG builds)—is that right?

robmaceachern (approved):
I'm not seeing `_isFirstResponderInDeletedSectionOrItem` being invoked in the `SystemFlowLayoutViewController` demo and the text field is still resigning first responder immediately. Is that expected?

Added a few comments on scoping the workaround more tightly, just to limit the blast radius if things went off the rails for some reason.

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -0,0 +1,348 @@
...
+/// doing its first responder-y things. This means this is the way to go. Not great, but not awful.
+///
+/// To accomplish this, we'll re-implement the broken bits of `_isFirstResponderInDeletedSectionOrItem`
+/// ourselves, and call back to the original implementation when we can.
+///
+extension ListView {
+    
+    // Note: If we need additional overrides, please subclass me, so we can
+    // wholesale delete this subclass when we drop iOS 16.4.
+    class IOS16_4_First_Responder_Bug_CollectionView : UICollectionView {
```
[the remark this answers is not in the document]
kyleve: I can do that!

kyleve:
> I'm not seeing _isFirstResponderInDeletedSectionOrItem being invoked in the SystemFlowLayoutViewController demo and the text field is still resigning first responder immediately. Is that expected?

This is expected, that uses a pure `UICollectionView` type, to test its behavior

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -40,7 +40,7 @@ public final class ListView : UIView
             behavior: self.behavior
         )
 
-        self.collectionView = UICollectionView(
+        self.collectionView = IOS16_4_First_Responder_Bug_CollectionView(
```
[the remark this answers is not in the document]
kyleve: I wanted to do it all in the workaround class so we can just delete it all at once.

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -0,0 +1,348 @@
...
+            
+            /// This is the Swift prototype of the function. There's no arguments, but ObjC
+            /// methods (when converted to their final C representation) have two implicit arguments,
+            /// `self`, the object, and `SEL`, the selector.
+            typealias SuperFunction = @convention(c) (AnyObject, Selector) -> Bool
+            
+            /// Make the pointer we got back into a Swift-callable function.
+            let super_function = unsafeBitCast(super_impl, to: SuperFunction.self)
+            
+
```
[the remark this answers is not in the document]
kyleve: Gotta let that code breathe, man 

(I'll fix)

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
kyleve: - [ ] Can we make this safer?

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -0,0 +1,348 @@
...
+///     }
+/// }
+/// ```
+///
+/// Hey, it works! Upon returning `false` from this method, the first responder remains
+/// doing its first responder-y things. This means this is the way to go. Not great, but not awful.
+///
+/// To accomplish this, we'll re-implement the broken bits of `_isFirstResponderInDeletedSectionOrItem`
+/// ourselves, and call back to the original implementation when we can.
+///
```
nononoah, line 119: Thanks for the thorough explanation of process here.

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
kyleve: We don't need access to those, because we own more of the state; so we can determine most of that information ourselves. Also just more generally, I wanted to poke at as little internal bits as possible; which means starting with the first responder, and then using our own info from therein.

zadr (approved):

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
kyleve: Ah true, we should also bail prior to 16.4.

- [ ] Bail before 16.4

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
zadr: we could look at the thread stack?

but would be safer to override performBatchUpdates, increment a flag (since, nested calls), call super and then decrement the flag?

Demo/Sources/Demos/Demo Screens/SystemFlowLayoutViewController.swift
```diff
@@ -99,16 +90,43 @@ extension SystemFlowLayoutViewController : UICollectionViewDataSource, UICollect
...
-            return collectionView.dequeueReusableSupplementaryView(ofKind: kind, withReuseIdentifier: "Footer", for: indexPath)
+            let header = collectionView.dequeueReusableSupplementaryView(
+                ofKind: kind,
+                withReuseIdentifier: "Header",
+                for: indexPath
+            ) as! FlowHeader
+            
+            header.textField.onFocus = { [weak collectionView] in
+                
+                /// ⚠️⚠️⚠️ Triggers the issue. ⚠️⚠️⚠️
```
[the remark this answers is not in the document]
kyleve: Are u triggered

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
kyleve: I'm going to leave this as is given the feature flag; and given these calls can be queued, we're not able to truly implement an allow/blacklist here.

PR #467, opened 2023-04-12
Prep 8.1.2

kyleve:
Noticed this one after merging. In theory it doesn't matter, but figure it's better to be safe

### Checklist


robmaceachern (approved):
Good catch.

PR #464, opened 2023-04-06
Allow swipe action view style to be set via environment

robmaceachern:
- Added `defaultSwipeActionsViewStyle` to `ListEnvironment`. This allows a `DefaultSwipeActionsView.Style` to be set on the environment when customizing the appearance of the default swipe action view.
- Added `styleEnvironmentKey` to the `ItemContentSwipeActionsView` protocol. 
- Added the `DefaultProviding` protocol for providing a default value for a type.
- The type of the `ItemContent.swipeActionsStyle` protocol requirement is now `SwipeActionsView.Style?` (previously `SwipeActionsView.Style`).
- `ItemContentSwipeActionsView.Style` must now conform to `DefaultProviding`.

This is a bit funky given how we allow each item to specify it's own `SwipeActionsView` type. You may want to start review at `PresentationState.ItemState.swift` to see how the style is resolved and work backwards from there. Open to other approaches, of course.

### Usage

```swift
list.environment.defaultSwipeActionsViewStyle = .init(
    containerCornerRadius: 6,
    equalButtonWidths: true,
    minWidth: 80
)
```

### Checklist

Please do the following before merging:

- [x] Ensure any public-facing changes are reflected in the [changelog](https://github.com/kyleve/Listable/blob/main/CHANGELOG.md). Include them in the `Main` section.

PR #468, opened 2023-04-17
Allow swipe action view style to be set via environment (w/o associated type complexities)

robmaceachern:
> **Note**
> This is targeting the `robmaceachern/feature/swipe-action-updates` feature branch.

> **Note**
> This is a simplified version of #464 that removes the `SwipeActionsView` associated type on `ItemContent`, as suggested [here](https://github.com/square/Listable/pull/464#discussion_r1160893506).

- Added `swipeActionsViewStyle` to `ListEnvironment`. This allows a `SwipeActionsView.Style` to be set on the list environment when customizing the appearance of the swipe action view.
- Removed the `SwipeActionsView` associated type on `ItemContent`.
- Removed the `ItemContentSwipeActionsView` protocol.
- Renamed `DefaultSwipeActionsView` to `SwipeActionsView`.
- The type of the `ItemContent.swipeActionsStyle` protocol requirement is now `SwipeActionsView.Style?` (previously `SwipeActionsView.Style`). When an item returns `nil` for this property the style set on the list environment will be used instead.

### Checklist


kyleve (approved):
Seems much cleaner, nice!

n8chur (approved):

ListableUI/Sources/Internal/PresentationState/PresentationState.ItemState.swift
```diff
@@ -313,7 +313,11 @@ extension PresentationState
                         
             // Apply Swipe To Action Appearance
             if let actions = self.model.swipeActions {
-                cell.contentContainer.registerSwipeActionsIfNeeded(actions: actions, style: model.content.swipeActionsStyle, reason: reason)
+                cell.contentContainer.registerSwipeActionsIfNeeded(
+                    actions: actions,
+                    style: model.content.swipeActionsStyle ?? environment[SwipeActionsViewStyleKey.self],
```
nononoah, line 318: Neatly and intuitively done!

ListableUI/Sources/Item/ItemContent.swift
```diff
@@ -352,21 +352,19 @@ public protocol ItemContent : AnyItemConvertible where Coordinator.ItemContentTy
...
-    
-    /// The view type to use to render swipe actions (delete, etc) for this content.
-    /// A default implementation, which matches `UITableView`, is provided.
-    associatedtype SwipeActionsView: ItemContentSwipeActionsView = DefaultSwipeActionsView
 
-    /// Specify a swipe action style for this content.
-    var swipeActionsStyle: SwipeActionsView.Style { get }
+    /// The swipe action style for this content.
+    ///
+    /// If this is `nil`, the style specified in the environment will be used.
```
nononoah, line 358: This isn't quite true, I think. If you fail to specify one in the environment, the default value for the associated key will be used, right?

nononoah (approved):

ListableUI/Sources/Item/ItemContent.swift
robmaceachern: Yeah maybe specified is the wrong term. How about "the style provided by the `SwipeActionsViewStyleKey` environment value will be used."?

PR #470, opened 2023-04-19
Fix versioning

kyleve:
`source cdn.cocoapods.org` doesn't seem to actually do anything... removing.


BlueprintUILists.podspec
```diff
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
   s.swift_versions = [LISTABLE_SWIFT_VERSION]
```
kyleve, line 14: Also cleaned up versioning consumption.

Podfile
```diff
@@ -1,5 +1,3 @@
-source 'https://cdn.cocoapods.org/'
```
kyleve, line 1: This is the actual fix

PR #471, opened 2023-04-19
Caffeinated layout contract adherence

watt:
Tweaks to the edge case measurement behavior of `List` elements in order to fully adhere to Blueprint's [Caffeinated Layout](https://github.com/square/Blueprint/pull/447). See the [Layout protocol documentation](https://square.github.io/Blueprint/Protocols.html#/s:11BlueprintUI6LayoutP) for full details, but the gist is that for a fixed width or height, the other dimension must grow as the size constraint grows (or rather, it cannot ever _shrink_).

This PR handles the edge cases of a filling List in an `unconstrained` constraint by returning `CGFloat.infinity` instead of `fatalError`, and ensures that the size at a constraint of `0` is not larger than the size at a constraint `> 0`.

### Checklist


BlueprintUILists/Sources/List.swift
```diff
@@ -241,6 +247,9 @@ extension List {
                         )
                     }
                 case .natural:
+                    if case .caffeinated = layoutMode, let maxHeight = constraint.height.constrainedValue {
+                        return min(size.contentSize.height, maxHeight)
```
watt, line 251: We could alternatively apply this clamping in `ListView.contentSize()`, but here it is scoped to only this layout mode.

kyleve: This seems fine to me!

kyleve (approved):

PR #473, opened 2023-04-20
Leading swipe actions

robmaceachern:
> **Note**
> This is targeting the `robmaceachern/feature/swipe-action-updates` feature branch.

This adds the ability to configure leading swipe actions to a cell.

- Add the ability to configure leading swipe actions on `Item`s.
- Renamed `Item.swipeActions` to `Item.leadingSwipeActions`.
- Renamed `DefaultItemProperties.swipeActions` to `leadingSwipeActions`.

### Discussion

I was able to mostly plug into the existing swipe action configuration with some tweaks depending on which side the actions are presented from. Most `SwipeActionState` cases have been augmented with a `Side` associated value in order to keep track of the state of the swipe actions and we essentially filter for "relevant" states when updating the individual swipe action views.

This will probably be fairly challenging in terms of code review so thanks in advance for taking a look (I suggest you use the split diff). I would also really appreciate if someone would boot up the demo in the simulator or device to see what I've broken.

### Demo

https://user-images.githubusercontent.com/430436/233661437-9c677dc9-dd31-40a7-930b-2acb0591c890.mov

### Checklist


ListableUI/Sources/Internal/DirectionalPanGestureRecognizer.swift
robmaceachern, line 1: This is adapted from the existing `LeftPanGestureRecognizer` to support both horizontal directions.

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -160,54 +217,46 @@ extension ItemCell {
...
                     if willPerformAction {
-                        swipeState = .willPerformFirstActionAutomatically
-                    } else if currentSwipeOffset > keepOpenOffset {
-                        swipeState = .open
+                        swipeState = .willPerformFirstActionAutomatically(side)
                     } else {
-                        swipeState = .closed
+                        swipeState = .open(side)
                     }
-
```
robmaceachern, line 195: This is the only (intentional) functional change I made. We were handling the `velocity == 0` case specially but in my testing I could never get the velocity to be exactly zero so I figured I'd just lump it in with the `isClosing` state to simplify things.

Let me know if you have any context for this path, and I can add it back if we want to be cautious.

ListableUI/Sources/Internal/SwipeActionsView.swift
```diff
@@ -10,6 +10,11 @@ import UIKit
 private let haptics = UIImpactFeedbackGenerator(style: .light)
 
 public final class SwipeActionsView: UIView {
```
robmaceachern, line 12: I'd like to make this internal now but the nested `Style` needs to be public. I think it'd be worth doing in a follow-up. What do you think?

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -85,51 +109,71 @@ extension ItemCell {
...
+            deregisterSwipeIfNeeded(for: .right)
+        }
+        
+        func registerLeadingSwipeActionsIfNeeded(actions: SwipeActionsConfiguration, style: SwipeActionsView.Style, reason: ApplyReason) {
+            registerSwipeActionsIfNeeded(side: .left, actions: actions, style: style, reason: reason)
+        }
+        
+        func registerTrailingSwipeActionsIfNeeded(actions: SwipeActionsConfiguration, style: SwipeActionsView.Style, reason: ApplyReason) {
+            registerSwipeActionsIfNeeded(side: .right, actions: actions, style: style, reason: reason)
+        }
```
robmaceachern, line 141: Note: I'm not actually checking the interface layout direction. The leading actions are plopped into the left and the trailing into the right side. If we want to support that in the future the client API shouldn't need to change, we'll just juggle the `side`s here and in deregistration. I figured that it was best to leave it out until it's actually needed.

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -63,19 +73,33 @@ extension ItemCell {
 
             case .expandActions:
 
-                xOriginOffset = -bounds.width
+                switch swipeViewSide {
+                case .left:
+                    xOriginOffset = bounds.width
+                case .right:
+                    xOriginOffset = -bounds.width
+                }
```
robmaceachern, line 81: Is there a more concise way to write these negations/tweaks? I end up doing similar things elsewhere.

ListableUI/Sources/Internal/SwipeActionsView.swift
n8chur: I'd be in favor of this in a follow up!

Demo/Sources/Demos/Demo Screens/SwipeActionsViewController.swift
```diff
@@ -156,6 +187,19 @@ final class SwipeActionsViewController: UIViewController  {
         }
         reloadData(animated: true)
     }
+    
+    private let shareURL = URL(string: "https://www.youtube.com/watch?v=dQw4w9WgXcQ")!
```
n8chur, line 191: 🎤🕺

n8chur (approved):
Looks good!

One thing I did notice which I don't believe to be a regression from these changes, is that you cannot re-grab the swipe view once actions are presented like you can with native implementations in Apple's app—as soon as you touch down on the cell the swipe actions are collapsed:

https://user-images.githubusercontent.com/671271/233689927-32be890f-e3e7-4616-8583-6f543468971f.mp4

Definitely doesn't need to be solved as part of this PR, but something we may want to follow up on or make a ticket for if it's not already known/intentional!

Demo/Sources/Demos/Demo Screens/SwipeActionsViewController.swift
kyleve: lol

Demo/Sources/Demos/Demo Screens/SwipeActionsViewController.swift
```diff
@@ -86,14 +89,42 @@ final class SwipeActionsViewController: UIViewController  {
...
+        
+        SwipeActionsConfiguration(performsFirstActionWithFullSwipe: true) {
+            SwipeAction(
+                title: nil,
+                accessibilityLabel: "Open Video",
+                backgroundColor: .systemBlue,
+                image: UIImage(systemName: "video.fill")
+            ) { [weak self] expandActions in
+                self?.open(item: item) {
+                    expandActions(false)
```
kyleve, line 110: This is reminding me I never replaced this with an enum that makes more sense than a bool. I'll add it to my backlog.

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -9,21 +9,25 @@ import UIKit
...
 
         let contentView : Content.ContentView
-
-        private var swipeConfiguration: SwipeConfiguration?
         
-        var swipeActionsView : SwipeActionsView? {
-            swipeConfiguration?.swipeView
+        private var configurations: [Side: SwipeConfiguration] = [:]
+        
+        private var swipeAccessibilityCustomActions: [Side: [AccessibilitySwipeAction]] = [:] {
```
kyleve, line 21: Rather than a dictionary, would a small struct with two properties be better, given in theory neither of these can be nil, just empty?

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
kyleve: You mean per side?

I tend to add [little helpers](https://github.com/squareup/market/blob/main/ios/MarketCore/Sources/Primitives/MarketSizeClass.swift#L85-L94) for things like that, though I know not everyone finds those as readable.

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
kyleve: Makes sense, I was going to ask about this. Agree we can deal with it later.

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
kyleve: I don't remember exactly. We could always tag KB since this is open source :trollface:

kyleve (approved):
This looks good, thanks!

One thing I would test / add to the demo, is to ensure that cell recycling is working properly. Eg, also add some rows that have leading/trailing items only and ensure that as cells are recycled during scroll, things work as expected.

robmaceachern:
> One thing I would test / add to the demo, is to ensure that cell recycling is working properly.

I've added a control in the demo that configures some rows without swipe actions (and some labels to view the config) and it all seems to be working properly in my testing.

<details>

9126771E-326B-457B-87D7-C6A93AD5BC6E_1_102_o

</details>

PR #474, opened 2023-04-24
Use accessibility label on swipe action if available

robmaceachern:
- Swipe actions will now use the `accessibilityLabel` (if non-nil) when the user is using assistive technologies such as VoiceOver.

### Checklist

### DIFF
--- a/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -331,7 +331,10 @@ private class AccessibilitySwipeAction: UIAccessibilityCustomAction {
     init(action: SwipeAction, side: Side, target: Any?, selector: Selector) {
         self.action = action
         self.side = side
-        super.init(name: action.title ?? "", target: target, selector: selector)
+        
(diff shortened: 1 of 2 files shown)


ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -331,7 +331,10 @@ private class AccessibilitySwipeAction: UIAccessibilityCustomAction {
     init(action: SwipeAction, side: Side, target: Any?, selector: Selector) {
         self.action = action
         self.side = side
-        super.init(name: action.title ?? "", target: target, selector: selector)
+        
+        let name = action.accessibilityLabel ?? action.title ?? ""
```
robmaceachern, line 335: The `"""` default really should never be hit since we have an assertion in `SwipeAction`:

```
        if title == nil || title?.isEmpty == true {
            precondition(
                accessibilityLabel?.isEmpty == false,
                "You must provide a title or an accessibilityLabel to a SwipeAction to ensure proper VoiceOver support."
            )
        }
```

But if you can think of a better fallback value we can plug it in instead.

kyleve (approved):
Nice find!

PR #479, opened 2023-04-26
Fix issue where first tap on a header with no pressed background would eat the tap

kyleve:
### Checklist


ListableUI/Sources/Internal/HeaderFooterContentView.swift
```diff
@@ -146,7 +146,7 @@ final class HeaderFooterContentView<Content:HeaderFooterContent> : UIView
             break
         
         case .began, .changed:
-            self.pressedBackground.alpha = 1.0
```
kyleve, line 149: This is the fix. Doh.

n8chur (approved):

ListableUI/Sources/Internal/HeaderFooterContentView.swift
kyleve: (Before, this was loading the view, which threw it into the view hierarchy, which would cancel the press gesture. This guarantees we don't do that anymore.)

PR #485, opened 2023-05-15
Added an underlay view to item cells

adamsocik1:
Added an optional underlay view to item cells. 

The problem mainly solved was to provide a view where we can set a background color when performing swipe actions with insets for cells that have a different color background than their list. In the before / after images, if you look at the swipe action, you'll notice that there is no longe a white "border" visible around the swipe action. The changes to the demo for the images are not a part of this PR, but clients will have to just implement something like:

```
        func underlayDecorationElement(with info: ApplyItemContentInfo) -> Element? {
            Empty().box(background: .cyan)
        }
```

|Before|After|
|:-:|:-:|
|image1|image2|

### Checklist

### DIFF
--- a/ListableUI/Sources/Internal/ItemCell.swift
+++ b/ListableUI/Sources/Internal/ItemCell.swift
@@ -42,6 +42,21 @@ final class ItemCell<Content:ItemContent> : UICollectionViewCell, AnyItemCell
     
     private(set) var overlayDecorationIfLoaded : OverlayDecorationView? = nil
     
+    private(set) lazy var underlayDecoration : OverlayDecorationView = {
+        let view = OverlayDecorationView(
+            content: Content.createReusableUnderlayDecorationView(frame:bounds),
+            frame: bounds
+        )
+        
+        self.underlayDecorationIfLoaded = view
+        
+        self.contentView.insertSubview(view, belowSubview: self.contentContainer)
+        
+        return view
+    }()
+    
+    private(set) var underlayDecorationIfLoaded : OverlayDecorationView? = nil
+    
     let contentContainer : ContentContainerView
 
     private(set) lazy var background : Content.BackgroundView = {
@@ -161,6 +176,7 @@ final class ItemCell<Content:ItemContent> : UICollectionViewCell, AnyItemCell
         self.contentContainer.frame = self.contentView.bounds
         
         self.overlayDecorationIfLoaded?.frame = self.contentView.bounds
+        self.underlayDecorationIfLoaded?.frame = self.contentView.bounds
     }
     
     // MARK: AnyItemCell
--- a/ListableUI/Sources/Internal/ItemCell.swift
+++ b/ListableUI/Sources/Internal/ItemCell.swift
@@ -42,8 +42,8 @@ final class ItemCell<Content:ItemContent> : UICollectionViewCell, AnyItemCell
     
     private(set) var overlayDecorationIfLoaded : OverlayDecorationView? = nil
     
-    private(set) lazy var underlayDecoration : OverlayDecorationView = {
-        let view = OverlayDecorationView(
+    private(set) lazy var underlayDecoration : UnderlayDecorationView = {
+        let view = UnderlayDecorationView(
             content: Content.createReusableUnderlayDecorationView(frame:bounds),
             frame: bounds
         )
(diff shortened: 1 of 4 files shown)


ListableUI/Sources/Item/ItemContent.swift
```diff
@@ -441,6 +441,14 @@ public protocol ItemContent : AnyItemConvertible where Coordinator.ItemContentTy
     /// do that work in `apply(to:)`, so the appearance will be updated if the appearance of content changes.
     static func createReusableOverlayDecorationView(frame : CGRect) -> OverlayDecorationView
     
+    /// Create and return a new underrlay decoration view used to render any required decorations under the content.
+    /// The decoration view appears under all content as a background, and is not affected by swipe actions.
+    ///
+    /// ## Note
+    /// Do not do configuration in this method that will be changed by your view's theme or appearance – instead
+    /// do that work in `apply(to:)`, so the appearance will be updated if the appearance of content changes.
+    static func createReusableUnderlayDecorationView(frame : CGRect) -> OverlayDecorationView
```
kyleve, line 450: I think we'll want a `associatedtype UnderlayDecorationView:UIView = UIView`, not reuse `OverlayDecorationView` here

kyleve (approved):
Nice! Thank you for doing this, looks great! Just one comment on not re-using the associated type from the overlay view.

ListableUI/Sources/Item/ItemContent.swift
adamsocik1: Updated! Was trying to avoid some duplication

ListableUI/Sources/Internal/ItemCell.swift
```diff
@@ -253,6 +269,45 @@ extension ItemCell {
             nil
         }
     }
+    
+    final class UnderlayDecorationView : UIView {
```
kyleve, line 273: Fwiw, you could avoid duplicating this bit, by doing: 

```swift
final class DecorationView<ContentView:UIView> : UIView {

    let content : ContentView

    init(content : ContentView, frame: CGRect) {

        self.content = content

        super.init(frame: frame)

        self.content.frame = bounds
        self.addSubview(self.content)

        self.isUserInteractionEnabled = false
    }

    @available(*, unavailable)
    required init?(coder: NSCoder) { fatalError() }

    override func layoutSubviews() {
        super.layoutSubviews()

        self.content.frame = self.bounds
    }

    override var isAccessibilityElement: Bool {
        get { false }
        set { fatalError("Cannot set isAccessibilityElement.") }
    }

    override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
        false
    }

    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
        nil
    }
}
```

And then doing `DecorationView<Content.UnderlayDecorationView>`, etc.

ListableUI/Sources/Item/ItemContent.swift
```diff
@@ -441,6 +444,14 @@ public protocol ItemContent : AnyItemConvertible where Coordinator.ItemContentTy
     /// do that work in `apply(to:)`, so the appearance will be updated if the appearance of content changes.
     static func createReusableOverlayDecorationView(frame : CGRect) -> OverlayDecorationView
     
+    /// Create and return a new underrlay decoration view used to render any required decorations under the content.
```
kyleve, line 447: Nit: `underrlay` -> `underlay`

ListableUI/Sources/Internal/ItemCell.swift
kyleve: (Eg, unifying this and the `OverlayDecorationView` in this same file)

ListableUI/Sources/Item/ItemContent.swift
kyleve: This isn't a big deal for Blueprint-land where everything is a BlueprintView, but for the UIKit-based base API, folks will likely want to provide differing views for both.

ListableUI/Sources/Internal/ItemCell.swift
adamsocik1: Good idea, updated!

ListableUI/Sources/Item/ItemContent.swift
adamsocik1: Ah, got it, makes sense

PR #489, opened 2023-05-19
Swipe Action Updates

kyleve:
A few changes I need for the delete button work in Market:

- Add `Apply*Info` to the `BlueprintEnvironment`: This allows us to magically wire up tapping the remove button exposes the swipe actions in the Market implementation: https://github.com/squareup/market/pull/6261
- Update some `SwipeAction` property and `typealias` names to align better with what each actually does.
- Replace the `true/false` flag passed to the swipe action completion with an enum, so it's clear from the callsite what the argument is actually for.

### Checklist


BlueprintUILists/Sources/BlueprintHeaderFooterContent.swift
```diff
@@ -90,6 +90,19 @@ where
 }
 
 
+extension Environment {
+        
+    public internal(set) var applyHeaderFooterContentInfo : ApplyHeaderFooterContentInfo? {
```
kyleve, line 95: - [ ] Docs

BlueprintUILists/Sources/BlueprintItemContent.swift
```diff
@@ -115,6 +115,20 @@ public protocol BlueprintItemContent : ItemContent
 }
 
 
+extension Environment {
+        
+    public internal(set) var applyItemContentInfo : ApplyItemContentInfo? {
```
kyleve, line 120: - [ ] Docs

BlueprintUILists/Sources/BlueprintHeaderFooterContent.swift
```diff
@@ -160,3 +175,14 @@ public extension BlueprintHeaderFooterContent
         return view
     }
 }
+
+
+fileprivate extension Element {
+    
+    func wrapInBlueprintEnvironment(with info : ApplyHeaderFooterContentInfo) -> Element {
```
nononoah, line 182: 🐴 

Perhaps `adaptedEnvironment(with info: ApplyHeaderFooterContentInfo)`? Perhaps `wrap` is the more conventional choice in the repo.

BlueprintUILists/Sources/BlueprintHeaderFooterContent.swift
```diff
@@ -160,3 +175,14 @@ public extension BlueprintHeaderFooterContent
...
+        self.adaptedEnvironment { env in
+            env = info.environment.blueprintEnvironment
+            env.applyHeaderFooterContentInfo = info
```
nononoah, line 185: I half-expected `info.environment.blueprintEnvironment`, since it's already been set on the info, to be configured as such.

Sanity check, do we need to worry about the `info -> env -> info` relationship at all?

ListableUI/Sources/Internal/PresentationState/PresentationState.ItemState.swift
```diff
@@ -281,11 +281,35 @@ extension PresentationState
...
+                    
+                    guard let swipeActions = self.model.leadingSwipeActions else {
+                        assertionFailure("Cannot showLeadingSwipeActions for `\(self.model.identifier)`, as no swipe actions have been provided.")
+                        return
+                    }
+                    
+                    guard swipeActions.actions.isEmpty == false else {
+                        assertionFailure("Cannot showLeadingSwipeActions for `\(self.model.identifier)`, as no swipe actions have been provided.")
+                        return
+                    }
```
nononoah, line 295: 🐴 

```suggestion
                    guard self.model.leadingSwipeActions?.isEmpty == false else {
                        assertionFailure("Cannot showLeadingSwipeActions for `\(self.model.identifier)`, as no swipe actions have been provided.")
                        return
                    }
```

ListableUI/Sources/Internal/PresentationState/PresentationState.ItemState.swift
```diff
@@ -281,11 +281,35 @@ extension PresentationState
...
+                    
+                    guard let swipeActions = self.model.trailingSwipeActions else {
+                        assertionFailure("Cannot showTrailingSwipeActions for `\(self.model.identifier)`, as no swipe actions have been provided.")
+                        return
+                    }
+                    
+                    guard swipeActions.actions.isEmpty == false else {
+                        assertionFailure("Cannot showTrailingSwipeActions for `\(self.model.identifier)`, as no swipe actions have been provided.")
+                        return
+                    }
```
nononoah, line 310: Same hobbyhorse here to take or leave.

ListableUI/Sources/SwipeActionsConfiguration.swift
```diff
@@ -55,10 +55,10 @@ public struct SwipeAction {
     ///
     /// Pass in `true` to expand the actions (typically only used when deleting the row)
     /// or `false` to collapse them.
-    public typealias CompletionHandler = (_ expandActions: Bool) -> Void
+    public typealias OnDidPerformAction = (OnDidPerformActionAnimation) -> Void
```
nononoah, line 58: Minor, but nice updates!

nononoah (approved):

BlueprintUILists/Sources/BlueprintHeaderFooterContent.swift
kyleve: I can change this; it's internal and only used here.

- [ ] Rename

BlueprintUILists/Sources/BlueprintHeaderFooterContent.swift
kyleve: > I half-expected info.environment.blueprintEnvironment, since it's already been set on the info, to be configured as such.

This is a bit of an odd case; we're intentionally replacing the entire environment in the `adaptedEnvironment` call, since items and headers/footers can exist in two modes: In the view hierarchy, and during measurement, which isn't in a view hierarchy. Thus, we want to always smash the default value with the one from the view.

> Sanity check, do we need to worry about the info -> env -> info relationship at all?

I don't _think_, so, but I can write a couple retain cycle tests to verify.

- [ ] Test retain cycles

ListableUI/Sources/Internal/PresentationState/PresentationState.ItemState.swift
kyleve: I started this, but explicitly split it into two asserts, so when hitting the breakpoint, you'd understand if there was a swipe actions config vs if the swipe actions array was empty.

kyleve:
Note, build is failing due to something around xrOS: https://github.com/XcodesOrg/XcodesApp/issues/401

PR #493, opened 2023-06-13
Paged list layouts now support using bounds, instead of itemInsets

kyleve:
This updates Paged List Layouts to respect and use the `bounds` property, which will allow us to properly apply market screen widths in Market. I've removed `itemInsets`, since `bounds.padding` does this instead.

### Checklist


ListableUI/Sources/Layout/Paged/PagedListLayout.swift
```diff
@@ -175,31 +172,77 @@ final class PagedListLayout : ListLayout
...
+            width: bounds.width,
+            alignment: .center
+        ))
+        
+        let itemPosition = itemWidth.position(
+            with: pageSize.width,
+            defaultWidth: bounds.width.clamp(pageSize.width)
+        )
+        
+        var lastMaxY : CGFloat = 0
```
robmaceachern, line 203: Looks like this was pre-existing but the name `lastMaxY` is a bit confusing for the horizontal direction.

robmaceachern (approved):

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
kyleve: I use width/height and x/y throughout these layouts, and then flip them based on direction; since I found it clearer than "axis/cross axis", which is what blueprint uses.

PR #497, opened 2023-07-25
CONV-1435: Custom keyboard adjustment mode

gilgtm:
- Adds a .custom KeyboardAdjustmentMode to fully customize inset behavior

https://github.com/square/Listable/assets/73715873/0cbe07cf-8612-4b97-afe7-2519a08e1995

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 - A new `isContentScrollable` property is added to `ListView` to determine if the content size is large enough that the list can be scrolled to a new position without springing back to it's original position.
 
-- A new `custom` case is added to `KeyboardAdjustmentMode` which allows the consumer to fully customize the inset behavior. `onKeyboardFrameWillChange` and `customScrollViewInsets` have been added to `ListView` and `updateScrollViewInsets` is now public. All of these can be utilized in conjunction to respond to the keyboard and fully control the insets. They are only availble to consumers utilizing `ListView` directly.
+- A new `custom` case is added to `KeyboardAdjustmentMode` which allows the consumer to fully customize the inset behavior. `onKeyboardFrameWillChange` and `customScrollViewInsets` have been added to `ListView` and `updateScrollViewInsets` is now public. All of these can be utilized in conjunction to respond to the keyboard and fully control the insets. For now, these are available through `ListView` only.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 - A new `isContentScrollable` property is added to `ListView` to determine if the content size is large enough that the list can be scrolled to a new position without springing back to it's original position.
 
+- A new `custom` case is added to `KeyboardAdjustmentMode` which allows the consumer to fully customize the inset behavior. `onKeyboardFrameWillChange` and `customScrollViewInsets` have been added to `ListView` and `updateScrollViewInsets` is now public. All of these can be utilized in conjunction to respond to the keyboard and fully control the insets. They are only availble to consumers utilizing `ListView` directly.
+
 ### Removed
 
 ### Changed
--- a/ListableUI/Sources/Behavior.swift
+++ b/ListableUI/Sources/Behavior.swift
@@ -80,6 +80,11 @@ extension Behavior
         
         /// The `contentInset` of the list is adjusted when the keyboard appears or disappears.
         case adjustsWhenVisible
+
+        /// Consumer calculates the edge insets and handles setting them
+        /// via the `ListView#customScrollViewInsets` callback. Only use this option
+        /// when managing the `ListView` directly.
+        case custom
     }
     
     
--- a/ListableUI/Sources/Behavior.swift
+++ b/ListableUI/Sources/Behavior.swift
@@ -82,7 +82,7 @@ extension Behavior
         case adjustsWhenVisible
 
         /// Consumer calculates the edge insets and handles setting them
-        /// via the `ListView#customScrollViewInsets` callback. Only use this option
+        /// via the ``ListView/customScrollViewInsets`` callback. Only use this option
         /// when managing the `ListView` directly.
         case custom
     }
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 - A new `isContentScrollable` property is added to `ListView` to determine if the content size is large enough that the list can be scrolled to a new position without springing back to it's original position.
 
+- A new `custom` case is added to `KeyboardAdjustmentMode` which allows the consumer to fully customize the inset behavior. `onKeyboardFrameWillChange` and `customScrollViewInsets` have been added to `ListView` and `updateScrollViewInsets` is now public. All of these can be utilized in conjunction to respond to the keyboard and fully control the insets. They are only availble to consumers utilizing `ListView` directly.
```
nononoah, line 11:
```suggestion
- A new `custom` case is added to `KeyboardAdjustmentMode` which allows the consumer to fully customize the inset behavior. `onKeyboardFrameWillChange` and `customScrollViewInsets` have been added to `ListView` and `updateScrollViewInsets` is now public. All of these can be utilized in conjunction to respond to the keyboard and fully control the insets. For now, these are available through `ListView` only.
```

ListableUI/Sources/Behavior.swift
```diff
@@ -80,6 +80,11 @@ extension Behavior
         
         /// The `contentInset` of the list is adjusted when the keyboard appears or disappears.
         case adjustsWhenVisible
+
+        /// Consumer calculates the edge insets and handles setting them
+        /// via the `ListView#customScrollViewInsets` callback. Only use this option
```
nononoah, line 85: Probably best to use the DocC standard annotation here.

```suggestion
        /// via the ``ListView/customScrollViewInsets`` callback. Only use this option
```

Fine to leave as is as well of course.

nononoah (approved):

CHANGELOG.md
gilgtm: - [ ] todo

ListableUI/Sources/Behavior.swift
gilgtm: good to know!

- [ ] todo

ListableUI/Sources/Behavior.swift
```diff
@@ -80,6 +80,11 @@ extension Behavior
...
+        /// when managing the `ListView` directly.
+        case custom
```
kyleve, line 87: Thought: Instead of `customScrollViewInsets` on `ListView`, should this take in an associated value? Eg:

```swift
case custom(() -> UIEdgeInsets)
```

That way you can't forget to provide `customScrollViewInsets` if `.custom` is provided.

This will break `Equatable`, but we may be able to just remove that. Not 100% sure how painful that is.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -314,8 +314,30 @@ public final class ListView : UIView
...
         
-    private func updateScrollViewInsets()
+    /// Callback for when the keyboard changes
+    public typealias KeyboardFrameWillChangeCallback = (
+        KeyboardCurrentFrameProvider,
+        (animationDuration: Double, options: UIView.AnimationOptions)
+    ) -> Void
+
+    /// Called whenever a keyboard change is detected
+    public var onKeyboardFrameWillChange: KeyboardFrameWillChangeCallback? = nil
```
kyleve, line 324: We should expose this on `ListProperties` as well (and stamp it onto the underlying list view in the `configure` call here.

Same goes for `customScrollViewInsets` below, but pending resolution of [this](https://github.com/square/Listable/pull/497/files#r1276977270) comment.

ListableUI/Sources/Behavior.swift
kyleve: (I forget if `Equatable` on this type was actually useful in any way, or I just added it to add it)

gilgtm: I tried that in my first attempt but I removed it because [we do an equality check](https://github.com/square/Listable/blob/main/ListableUI/Sources/ListView/ListView.swift#L238-L246) every time the `behavior` property is set.

ListableUI/Sources/ListView/ListView.swift
gilgtm: - [ ] todo

gilgtm: I added `onKeyboardFrameWillChange` to `ListProperties` but I'm not sure what to do with `customScrollViewInsets`.  Please see my [comment](https://github.com/square/Listable/pull/497#discussion_r1277934131).

kyleve: How expensive is the calculation done in `customScrollViewInsets`? Is it expensive to just re-call it on every set of the closure, and then only apply the resulting inset value(s) to the collection view if they changed?

gilgtm: It's not expensive. In the demo app it just returns a pre-calculated value that's stored by the consumer. 

Currently, changing `customScrollViewInsets` has no effect until the consumer calls `updateScrollViewInsets`.

>  Is it expensive to just re-call it on every set of the closure, and then only apply the resulting inset value(s) to the collection view if they changed?

I think I lost the thread here, do you mind being more explicit about what you're suggesting. I'm not sure if this is related to moving it to behavior?

gilgtm: I thought about this some more and I think what you're suggesting is that we add an associated value to `.custom` (as you mentioned) above, and always consider the `.custom` case to be non-equal. I don't think there would be any perf concerns related specifically to the function changing, since we don't actually call it. But (without additional logic) it will cause `applyBehavior` to be called every time.

gilgtm: I'm going to punt on this for now. Let me know if we should revisit.

kyleve: Sound good!

PR #500, opened 2023-08-03
CONV-1435: Add scroll indicator insets to customScrollViewInsets

gilgtm:
- Before: customScrollViewInsets only returns insets to adjust contentInset
- After: customScrollViewInsets function signature matches calculateScrollViewInsets so you can specify scroll-indicator insets
- Why? In the chat demo, the scroll bar was disappearing below the bottom inset of the list. It's likely that if you are modifying contentInset you will have to modify scroll indicator insets as well

### Before

https://github.com/square/Listable/assets/73715873/ef74979d-83fb-478d-a3df-0a33da9f6148

### After

https://github.com/square/Listable/assets/73715873/f4e048b1-6904-4ffb-a5b1-e836e18caf5d

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -341,18 +341,15 @@ public final class ListView : UIView
     /// whenever insets require an update.
     public func updateScrollViewInsets()
     {
+        let insets: (content: UIEdgeInsets, horizontalScroll: UIEdgeInsets, verticalScroll: UIEdgeInsets)
         if case .custom = self.behavior.keyboardAdjustmentMode {
-            let insets = self.customScrollViewInsets()
-            self.collectionView.contentInset = insets.content
-            self.collectionView.horizontalScrollIndicatorInsets = insets.horizontalScroll
-            self.collectionView.verticalScrollIndicatorInsets = insets.verticalScroll
-            return
+            insets = self.customScrollViewInsets()
+        } else {
+            insets = self.calculateScrollViewInsets(
+                with: self.keyboardObserver.currentFrame(in: self)
+            )
         }
 
-        let insets = self.calculateScrollViewInsets(
-            with: self.keyboardObserver.currentFrame(in: self)
-        )
-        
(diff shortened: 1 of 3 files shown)


ListableUI/Sources/ListView/ListView.swift
```diff
@@ -329,7 +329,11 @@ public final class ListView : UIView
 
     /// This callback determines the scroll view's insets only when
     /// `behavior.keyboardAdjustmentMode` is `.custom`
-    public var customScrollViewInsets: () -> UIEdgeInsets = { .zero }
+    public var customScrollViewInsets: () -> (
```
kyleve, line 332: Up to you, but it may be nice to make this a real struct vs just a tuple, so we can document it, etc. But just a suggestion, not a big deal.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -338,7 +342,10 @@ public final class ListView : UIView
     public func updateScrollViewInsets()
     {
         if case .custom = self.behavior.keyboardAdjustmentMode {
-            self.collectionView.contentInset = self.customScrollViewInsets()
+            let insets = self.customScrollViewInsets()
+            self.collectionView.contentInset = insets.content
```
kyleve, line 346: Nit: We should wrap each of these assignments in a `!=` check (as is done below) before assigning: It's mostly a "belt and suspenders" thing, but sometimes, even setting these values to the same exact value as before can break in-progress scroll events for whatever reason.

kyleve (approved):

ListableUI/Sources/ListView/ListView.swift
gilgtm: Added! new type `ScrollViewInsets`

ListableUI/Sources/ListView/ListView.swift
gilgtm: Refactored!

kyleve: 🙏 Thank you!

PR #506, opened 2023-09-14
Update workaround versions

kyleve:
Fix messaging for version workarounds; this is still broken in 17 RC


robmaceachern (approved):
Will the madness ever end?

PR #507, opened 2023-09-19
Fix supplementary view + contained first responder reuse issue

kyleve:
The below works around a (seeming?) bug or odd behavior in `UICollectionView`, where it tries to be smart about recycling supplementary views that contain a first responder such as a text field. Specifically, it holds onto a supplementary view that contains a first responder, not immediately recycling it when it is scrolled out of view. That ensures that the keyboard isn't immediately dismissed, which would be jarring.

...Unfortunately, this doesn't seem to actually work in practice very well. When the supplementary view  is scrolled back _into_ view, and we're asked to dequeue a view, the collection view hands us back a _different_ view, leading to double views that get stacked on top of each other in the layout, leading to a bunch of weirdness.

So, to work around this, we do a few things:

1) We begin tracking which supplementary views currently contain a first responder. For practicality of implementation, we only track text fields right now. This could change, but is harder, given there's no generic "first responder changed" notification. This code lives in `ListView`.

2) We update `ListLayoutContent.content(in: ...)` to _always_ return supplementary info when a supplementary view contains a first responder, even when out of frame. This ensures the supplementary view instance is kept alive by the collection view. You can see that happening here:

image

3) Within `collectionView(_:viewForSupplementaryElementOfKind:at:)`, we check to see if there's a live, existing `visibleContainer` (aka the supplementary view) view, and if there is, we return _that_, instead of just dequeuing a new, wrong view.

After all that, the correct thing happens.

### Checklist

### DIFF
--- a/ListableUI/Sources/Internal/UIView.swift
+++ b/ListableUI/Sources/Internal/UIView.swift
@@ -30,10 +30,4 @@ extension UIView {
         
         return nil
     }
-    
-    func isInside(superview : UIView) -> Bool {
-        
-        sequence(first: self, next: \.superview)
-            .contains(superview)
-    }
 }
--- a/ListableUI/Sources/Internal/UIView.swift
+++ b/ListableUI/Sources/Internal/UIView.swift
@@ -30,4 +30,10 @@ extension UIView {
         
         return nil
     }
+    
+    func isInside(superview : UIView) -> Bool {
+        
+        sequence(first: self, next: \.superview)
+            .contains(superview)
+    }
 }
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -106,9 +106,11 @@ public final class ListView : UIView
         self.applyBehavior()
         self.updateScrollViewInsets()
         
-        // We register for first responder notifications so we can
-        // avoid recycling supplementary views that contain text fields,
-        // as this somehow breaks cell recycling.
+        /// We track first responder status in supplementary views
+        /// to fix a view recycling issue.
+        ///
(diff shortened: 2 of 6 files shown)


ListableUI/Sources/ListView/ListView.swift
robmaceachern, line 1: > For practicality of implementation, we only track text fields right now. This could change, but is harder, given there's no generic "first responder changed" notification.

If we _really_ need to, we can try out the `UIWindowFirstResponderDidChangeNotification` [private api](https://twitter.com/steipete/status/331774953336745985?lang=en). I think scoping to text fields is totally fine though.

robmaceachern (approved):
LGTM.

Do you happen to have a vanilla UICollectionView repro of this problem? I can file a feedback with Apple if you haven't already.

kyleve:
> Do you happen to have a vanilla UICollectionView repro of this problem? I can file a feedback with Apple if you haven't already.

I haven't, I couldn't repro in Flow layout which makes me think its either doing something similarly special, or I didn't quite poke things right.

ListableUI/Sources/Internal/UIView.swift
```diff
@@ -30,4 +30,10 @@ extension UIView {
         
         return nil
     }
+    
+    func isInside(superview : UIView) -> Bool {
+        
+        sequence(first: self, next: \.superview)
+            .contains(superview)
+    }
```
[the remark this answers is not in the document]
kyleve: Oops, yep!

- [ ] Remove

ListableUI/Sources/ListView/ListView.swift
kyleve: of course it's from Peter Steinberger

PR #510, opened 2023-10-16
[WIP DNR] Fix cross-axis natural measurements

kyleve:

### Checklist

### DIFF
--- a/BlueprintUILists/Sources/List.swift
+++ b/BlueprintUILists/Sources/List.swift
@@ -219,6 +219,7 @@ extension List {
             }()
             
             let height : CGFloat = {
+                
                 switch verticalFill {
                 case .fillParent:
                     if let max = constraint.height.constrainedValue {
@@ -250,7 +251,13 @@ extension List {
                     if case .caffeinated = layoutMode, let maxHeight = constraint.height.constrainedValue {
                         return min(size.contentSize.height, maxHeight)
                     }
-                    return size.contentSize.height
+                    
+                    /// Temporary
+                    if size.layoutDirection == .horizontal {
+                        return size.naturalWidth ?? size.contentSize.height
+                    } else {
+                        return size.contentSize.height
+                    }
                 }
             }()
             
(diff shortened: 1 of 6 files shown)


BlueprintUILists/Tests/List.MeasurementTests.swift
kyleve, line 1: - [ ] Can revert these changes

BlueprintUILists/Tests/XCTestCaseAdditions.swift
```diff
@@ -34,4 +34,9 @@ extension XCTestCase
         
         print("Iterations: \(iterations), Average Time: \(average)")
     }
+    
+    func testcase(_ name : String = "", _ block : () -> ())
```
kyleve, line 38: Copied from main lib

kyleve:
Still looking for reviews on this!

BlueprintUILists/Sources/List.swift
```diff
@@ -185,72 +185,50 @@ extension List {
             verticalFill : Measurement.FillRule
         ) -> CGSize
         {
+            precondition(
+                layoutMode == .caffeinated,
+                "Listable only supports the `.caffeinated` layout mode in Blueprint."
+            )
```
n8chur, line 191: Isn't this still feature flagged? Does this need to occur in this PR?

I think I'd prefer that the switch to only supporting `caffeinated` layout be in a separate PR that's combined with the work to do so in Market/Register if we can...

kyleve: The flag still exists, though in theory its 100% rolled out. I had been hoping maybe we'd be ripping out the old layout mode before this merged, but it looks like that's not happening! I'll add this back in.

n8chur: OK feel free to ping me on Slack for another review when it's ready!

kyleve: I also decided to take a swing at removing legacy layout: https://github.com/square/Blueprint/pull/475

PR #512, opened 2023-11-20
Remove override of performBatchUpdates, it causes warnings for consumers

kyleve:
There doesn't seem to be a way to represent `NS_NOESCAPE` in Swift here... Just going to remove this override since it's an entirely internal type anyway.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
+++ b/ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
@@ -134,11 +134,6 @@ extension ListView {
         
         required init?(coder: NSCoder) { fatalError() }
         
-        @available(*, unavailable, message: "Please use performBatchUpdates:changes:completion:.")
-        override func performBatchUpdates(_ updates: (() -> Void)?, completion: ((Bool) -> Void)? = nil) {
-            super.performBatchUpdates(updates, completion: completion)
-        }
-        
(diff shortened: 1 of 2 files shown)


ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
```diff
@@ -134,11 +134,6 @@ extension ListView {
         
         required init?(coder: NSCoder) { fatalError() }
         
-        @available(*, unavailable, message: "Please use performBatchUpdates:changes:completion:.")
-        override func performBatchUpdates(_ updates: (() -> Void)?, completion: ((Bool) -> Void)? = nil) {
-            super.performBatchUpdates(updates, completion: completion)
-        }
-        
         func performBatchUpdates(
```
robmaceachern, line 137: Minor: It might be worth adding some doc comments to this method since the normal `performBatchUpdates` will now show up in things like autocomplete results (and might help avoid any accidental invocation).

robmaceachern (approved):

ListableUI/Sources/ListView/ListView+iOS16.4Workaround.swift
kyleve: I'm not super worried about this; this method is only called in one place in the entire codebase (by design).

PR #516, opened 2023-12-05
Update swipe actions to more closely match iOS behavior

nsillik:
Update the swipe action interactions to better match the built in iOS behavior.

- If the swipe actions are open, and the user pans inside that cell you'll either grow the swipe actions (and possibly activate the default action) or interactively close the swipe actions.
- Touching anywhere outside of the cell with open swipe actions will close the swipe actions
- _Tapping_ inside of a cell with currently open cell will close the swipe actions.

https://github.com/square/Listable/assets/141675/780f07cf-7796-4fc2-bad8-e13a8b0e9087

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Changed
 
-- Update the swipe action interactions to more closely match iOS behavior
+- Update the swipe action interactions to more closely match iOS behavior.
 
 ### Misc
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Changed
 
+- Update the swipe action interactions to more closely match iOS behavior
+
 ### Misc
 
 ### Internal
--- a/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -118,10 +118,6 @@ extension ItemCell {
             }
         }
         
-        func isTouchWithinSwipeActionView(touch: UITouch) -> Bool {
-            configurations.values.first { $0.swipeView.contains(touch: touch) } != nil
-        }
-
         // MARK: - Swipe Registration
         
         func deregisterLeadingSwipeIfNeeded() {
@@ -286,7 +282,6 @@ extension ItemCell {
         }
 
         private func set(state: SwipeActionState, animated: Bool = false) {
-
             swipeState = state
 
             if animated {
--- a/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -161,6 +161,9 @@ extension ItemCell {
 
                 let panGestureRecognizer = DirectionalPanGestureRecognizer(direction: side.gestureDirection, target: self, action: #selector(handlePan))
                 addGestureRecognizer(panGestureRecognizer)
+                let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleTap))
+                tapGestureRecognizer.require(toFail: panGestureRecognizer)
+                addGestureRecognizer(tapGestureRecognizer)
 
                 configurations[side] = SwipeConfiguration(
                     panGestureRecognizer: panGestureRecognizer,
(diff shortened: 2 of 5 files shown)


ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1037,7 +1037,7 @@ public final class ListView : UIView
         
         guard let cell = self.liveCells.activeSwipeCell else { return false }
         
-        return cell.isTouchWithinSwipeActionView(touch: touch) == false
+        return cell.contains(touch: touch) == false
```
nsillik, line 1040: If the user is touching down _anywhere_ in the `activeSwipeCell` the `activeSwipeCell` will handle it.

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -182,6 +185,10 @@ extension ItemCell {
 
         private weak var listView : ListView? = nil
 
+        @objc private func handleTap(sender: UITapGestureRecognizer) {
+            set(state: .closed, animated: true)
+        }
```
nsillik, line 190: if we're tapped (and we're not panning) just close

kyleve:
Does this also handle the case where if there's an open swipe action on one row, swiping on another row will dismiss the previous row's actions and open the actions on the new row?

https://github.com/square/Listable/assets/327847/a7299666-ff8a-4bf8-bcfe-82dae55c28f7

nsillik:
it _does_ dismiss the previous row's actions, but it doesn't open the actions on the new one, i can take a swing at making that work though (i should also likely add the comments that i made above as code comments too)

kyleve:
> it _does_ dismiss the previous row's actions, but it doesn't open the actions on the new one, i can take a swing at making that work though

If you can figure out how to make it work, that _seems_ to be the last parity difference between this implementation and what iOS does.

> (i should also likely add the comments that i made above as code comments too)

+100

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -210,8 +218,12 @@ extension ItemCell {
                 && configuration.performsFirstActionWithFullSwipe
 
             if sender.state == .began {
-                self.listView?.liveCells.perform {
```
kyleve, line 213: Was losing `$0.closeSwipeActions()` intentional?

nsillik:
okay, got that last piece working, demo video updated

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Changed
 
+- Update the swipe action interactions to more closely match iOS behavior
```
robmaceachern, line 11: Micro nit:

```suggestion
- Update the swipe action interactions to more closely match iOS behavior.
```

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
robmaceachern, line 1: It's pretty subtle, but one thing that would be nice to support is the case where the swipe actions are open and the user pans back to close them. It's still not an interactive interaction when doing that directly. Not a huge deal but it'd be a nice touch.

https://github.com/square/Listable/assets/430436/4c51a234-93fa-4949-b2f5-8e1413dcdd24

robmaceachern (approved):
Changes LGTM but there's one micro interaction I wouldn't mind if we could support while we're here.

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
nsillik: oh man, yeah, this is absolutely a good call, but after some digging i think this is gonna be a much bigger refactor. i'm going to punt for now.

nsillik:
> > it _does_ dismiss the previous row's actions, but it doesn't open the actions on the new one, i can take a swing at making that work though
> 
> If you can figure out how to make it work, that _seems_ to be the last parity difference between this implementation and what iOS does.
> 
> > (i should also likely add the comments that i made above as code comments too)
> 
> +100

weird enough, while working on #519 i was going to add this interaction in, but it looks like it's working now:


https://github.com/square/Listable/assets/141675/f3e9a791-5869-440e-bebf-dce0715e0b5e

PR #517, opened 2023-12-08
Don't animate when dequeuing cells

kyleve:
https://block.atlassian.net/browse/UI-5081

We now force a non-animated layout before dequeuing a cell, to ensure that any implicit animations that may occur are avoided. This avoids issues where you may see an animation as a reused cell's changed state is applied.

### Checklist

- [x] POS Integration Test
- [x] Market Integration Test


n8chur (approved):
Seems reasonable, but this won't cause an _extra_ layout, would it? it's just going to happen earlier and non-animated since the cell should be `needsLayout = false` after this?

kyleve:
> Seems reasonable, but this won't cause an _extra_ layout, would it? it's just going to happen earlier and non-animated since the cell should be `needsLayout = false` after this?

I logged calls to layout and the number of calls seemed identical from what I could tell.

kyleve:
This is for some reason breaking the KIF tests, seemingly related to market search bar, will need to dig into that before merging: https://square-console.sqprod.co/bbr/production/ci_jobs/PR+ios+WK-SK4+Bazel-Pipeline-PR/3

CHANGELOG.md
kyleve, line 1: - [ ] A possible bug I found, should fix before merging: `ListView.VisibleContent.update` should assign 

```swift
            self.items = newItems
            self.headerFooters = newHeaderFooters
```

Before callbacks

PR #529, opened 2024-02-29
Bumping versions to 15.0.2

nsillik:
# [14.0.2] - 2024-02-29

### Fixed

- Fixed a bug that resulted in some swipe action buttons not being tappable.


robmaceachern (approved):
LGTM.

I'm going to see if there's a way to suppress the dates in the footers so that we don't always get these noisy doc diffs.

PR #530, opened 2024-02-29
chore: Generated documentation now uses a static copyright notice to avoid noisy diffs

robmaceachern:
Generated documentation now uses a static copyright notice to avoid noisy diffs.

In the past, running `Scripts/generate_docs.sh` would generate a huge diff as every page footer would update the timestamp to the latest day (example: https://github.com/square/Listable/pull/529).

This should make it much easier to diff real doc changes and prevent a lot of noise from showing up.

### Checklist

- [x] Wait for 15.0.2 release to land before merging (this is currently targeting that branch to avoid merge conflicts).


Scripts/generate_docs.sh
robmaceachern, line 1: The actual changes are in this file.

docs/BlueprintLists/Enums.html
```diff
@@ -185,7 +185,7 @@ <h4>Declaration</h4>
       </article>
     </div>
     <section class="footer">
-      <p>&copy; 2024 <a class="link" href="https://github.com/kyleve/Listable" target="_blank" rel="external noopener">Kyle</a>. All rights reserved. (Last updated: 2024-02-29)</p>
+      <p>&copy; Square, Inc. All rights reserved.</p>
       <p>Generated by <a class="link" href="https://github.com/realm/jazzy" target="_blank" rel="external noopener">jazzy ♪♫ v0.14.4</a>, a <a class="link" href="https://realm.io" target="_blank" rel="external noopener">Realm</a> project.</p>
```
robmaceachern, line 189: Note: The jazzy version (`jazzy ♪♫ v0.14.4`) is still dynamic and will cause mass updates any time Jazzy is updated but that should be relatively infrequent (a version bump may cause other changes anyways).

It looks like we could pass in a `[JAZZY_FAKE_VERSION](https://github.com/realm/jazzy/blob/a04e99e820d79ea0c3a8dd1d5298ef459a92fa46/lib/jazzy/doc.rb#L30C12-L30C30)` environment var to make it static, if we really wanted to.

Scripts/generate_docs.sh
kyleve: Ohhhh, thank you so much! I always wondered how to fix this...

PR #531, opened 2024-03-03
Fix SPOS reordering crash / header reuse issue + prep 14.0.3

kyleve:
https://block.atlassian.net/browse/UI-5594

Fix a crash that could occur during cell reuse if a list contained different types of headers. The wrong `ObjectIdentifier` was being compared and stored. I'm surprised this ever worked – but it seems like _some_ timing changed (maybe when we dropped iOS 14?) that this wasn't hit nearly as frequently – it's now pathological for any list with different header types.

### Checklist


ListableUI/Sources/ListView/ListView.Delegate.swift
kyleve, line 1: Actual fix is in here

CHANGELOG.md
```diff
@@ -1020,7 +1028,11 @@ listActions.scrolling.scrollToSection(
 Earlier releases were ad-hoc and not tracked. To see all changes, please reference [closed PRs on Github](https://github.com/kyleve/Listable/pulls?q=is%3Apr+is%3Aclosed).
 
 
-[Main]: https://github.com/square/Listable/compare/13.1.0...main
+[Main]: https://github.com/square/Listable/compare/14.0.3...main
+[14.0.3]: https://github.com/square/Listable/square/14.0.2...14.0.3
+[14.0.2]: https://github.com/square/Listable/square/14.0.1...14.0.2
```
kyleve, line 1033: These weren't added in the prior releases

ListableUI/Sources/ListView/ListView.Delegate.swift
kyleve: (Note: I'm gonna follow up with an automated test for this, but want to get this landed before branch cut.)

robmaceachern (approved):

sameerg-bit (approved):
Thanks for the fix!

PR #533, opened 2024-03-05
reorder control now proxies accessibility into a seperate element

RoyalPineapple:
This allows the gesture recognizer to be applied to a view without overriding its accessibility. 
Useful in situations where you'd like the gesture to apply to the entire cell for example.    

### Checklist

### DIFF
--- a/BlueprintUILists/Sources/ListReorderGesture.swift
+++ b/BlueprintUILists/Sources/ListReorderGesture.swift
@@ -5,6 +5,7 @@
 //  Created by Kyle Van Essen on 11/14/19.
 //
 
+import Accessibility
 import BlueprintUI
 import ListableUI
 import UIKit
@@ -38,6 +39,7 @@ import UIKit
 /// ```
 public struct ListReorderGesture : Element
 {
+    
     public enum Begins {
         case onTap
         case onLongPress
@@ -54,9 +56,8 @@ public struct ListReorderGesture : Element
     
     let actions : ReorderingActions
     
-    /// The acccessibility Label of the item that will be reordered.
-    /// This will be set as the gesture's accessibilityValue to provide a richer VoiceOver utterance.
-    public var reorderItemAccessibilityLabel : String? = nil
+    /// The acccessibility label for the reorder element. Defaults  to "Reorder".
+    public var accessibilityLabel : String?
     
     /// Creates a new re-order gesture which wraps the provided element.
     /// 
@@ -66,6 +67,7 @@ public struct ListReorderGesture : Element
         isEnabled : Bool = true,
         actions : ReorderingActions,
         begins: Begins = .onTap,
+        accessibilityLabel: String? = nil,
         wrapping element : Element
     ) {
         self.isEnabled =  isEnabled
@@ -73,6 +75,8 @@ public struct ListReorderGesture : Element
         self.actions = actions
 
         self.begins = begins
+        
+        self.accessibilityLabel = accessibilityLabel
         
         self.element = element
     }
@@ -88,24 +92,16 @@ public struct ListReorderGesture : Element
     public func backingViewDescription(with context: ViewDescriptionContext) -> ViewDescription?
     {
         return ViewDescription(View.self) { config in
+
             config.builder = {
                 View(frame: context.bounds, wrapping: self)
             }
+            config.contentView = { $0.containerView }
             
             config.apply { view in
-                view.isAccessibilityElement = true
-                view.accessibilityLabel = ListableLocalizedStrings.ReorderGesture.accessibilityLabel
-                view.accessibilityValue = reorderItemAccessibilityLabel
-                view.accessibilityHint = ListableLocalizedStrings.ReorderGesture.accessibilityHint
-                view.accessibilityTraits.formUnion(.button)
-                view.accessibilityCustomActions = accessibilityActions()
-                
-                view.recognizer.isEnabled = self.isEnabled
-                
-                view.recognizer.apply(actions: self.actions)
-                
-                view.recognizer.minimumPressDuration = begins == .onLongPress ? 0.5 : 0.0
+                view.apply(self)
             }
+            
         }
     }
 
@@ -118,9 +114,14 @@ public extension Element
     func listReorderGesture(
         with actions : ReorderingActions,
         isEnabled : Bool = true,
-        begins: ListReorderGesture.Begins = .onTap
+        begins: ListReorderGesture.Begins = .onTap,
+        accessibilityLabel: String? = nil
     ) -> Element {
-        ListReorderGesture(isEnabled: isEnabled, actions: actions, begins: begins, wrapping: self)
+        ListReorderGesture(isEnabled: isEnabled,
+                           actions: actions,
+                           begins: begins,
+                           accessibilityLabel: accessibilityLabel,
+                           wrapping: self)
     }
 }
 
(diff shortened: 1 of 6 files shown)


BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -5,6 +5,7 @@
 //  Created by Kyle Van Essen on 11/14/19.
 //
 
+import Accessibility
```
kyleve, line 8: huh, TIL

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -55,7 +56,7 @@ public struct ListReorderGesture : Element
     let actions : ReorderingActions
     
     /// The acccessibility Label of the item that will be reordered.
-    /// This will be set as the gesture's accessibilityValue to provide a richer VoiceOver utterance.
+    /// This will be used to provide a richer VoiceOver utterance useful for disambiguation as to which element will br reordered.,
```
kyleve, line 59: `br` -> `be`?

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -82,7 +86,7 @@ public struct ListReorderGesture : Element
     //
     
     public var content: ElementContent {
-        ElementContent(child: self.element)
+        ElementContent(measuring: self.element)
```
kyleve, line 89: Do we need to do this dance with the extra `BlueprintView`, or can you just add a content container view for Blueprint to add the elements to? Eg, `ScrollView` leverages this: https://github.com/square/Blueprint/blob/main/BlueprintUICommonControls/Sources/ScrollView.swift#L58

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -129,25 +127,91 @@ fileprivate extension ListReorderGesture
...
+        }
         
         init(frame: CGRect, wrapping : ListReorderGesture)
         {
             self.recognizer = .init()
             
             super.init(frame: frame)
-            
+            recognizer.accessibilityProxy = proxyElement
+            NotificationCenter.default.addObserver(forName: UIAccessibility.voiceOverStatusDidChangeNotification, object: nil, queue: nil) { [weak self] notification in
```
kyleve, line 148: Doesn't this return a token you need to retain and de-register in `deinit`?

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -129,25 +127,91 @@ fileprivate extension ListReorderGesture
 {
     private final class View : UIView
     {
+        
+        let blueprintView = BlueprintView()
         let recognizer : ItemReordering.GestureRecognizer
+        private lazy var proxyElement = UIAccessibilityElement(accessibilityContainer: self)
```
kyleve, line 133: Huh, TIL, I had no idea you could do this!

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -129,25 +127,91 @@ fileprivate extension ListReorderGesture
...
         init(frame: CGRect, wrapping : ListReorderGesture)
         {
             self.recognizer = .init()
             
             super.init(frame: frame)
-            
+            recognizer.accessibilityProxy = proxyElement
+            NotificationCenter.default.addObserver(forName: UIAccessibility.voiceOverStatusDidChangeNotification, object: nil, queue: nil) { [weak self] notification in
+                guard let self = self  else { return }
+                self.recognizer.minimumPressDuration = UIAccessibility.isVoiceOverRunning ? 0.0 : self.minimumPressDuration
```
kyleve, line 150: Can we centralize this logic and the logic in `didSet` by extracting it into a private func both call?

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -129,25 +127,91 @@ fileprivate extension ListReorderGesture
...
+        }
+        
+        override func layoutSubviews() {
+            super.layoutSubviews()
+            blueprintView.frame = bounds
+        }
+        
+        override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+            if UIAccessibility.isVoiceOverRunning,
+            UIAccessibility.focusedElement(using: nil) as? NSObject == proxyElement {
```
kyleve, line 197: Do you want to pass `.notificationVoiceOver` here? From the header:
```
// If no argument is used, the function will returned the element that was most recently focused.
```
Eg, not sure we care about `.notificationSwitchControl`?

BlueprintUILists/Sources/ListReorderGesture.swift
RoyalPineapple: Yes, because I return it in the `accessibilityElemenrts` override

RoyalPineapple: Ah yes, of course

BlueprintUILists/Sources/ListReorderGesture.swift
RoyalPineapple: ah, cool that works fine.

BlueprintUILists/Sources/ListReorderGesture.swift
RoyalPineapple: I was relying on the above check for `isVoiceOverRunning` but in the case where a switch is active we do only care about voiceover here so it does make sense.

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -129,25 +130,97 @@ fileprivate extension ListReorderGesture
...
+             }
         }
         
         @available(*, unavailable)
         required init?(coder aDecoder: NSCoder) {
             listableInternalFatal()
         }
+        
+        func apply(_ model: ListReorderGesture) {
+            if let itemName = model.reorderItemAccessibilityLabel {
```
kyleve, line 183: It doesn't look like we're using `reorderItemAccessibilityLabel` here?

ListableUI/Sources/Item/ItemReordering.swift
```diff
@@ -206,6 +210,7 @@ extension ItemReordering {
 
         @objc private func updated()
         {
+            guard accessibilityShouldContinue() else { return }
```
kyleve, line 213: Don't we need to set `self.state = .canceled` in these cases?

kyleve (approved):

BlueprintUILists/Sources/ListReorderGesture.swift
RoyalPineapple: `itemName` is used in the format string below.

kyleve: Oh doh

ListableUI/Sources/Item/ItemReordering.swift
RoyalPineapple: seems smart

BlueprintUILists/Sources/ListReorderGesture.swift
meherkasam-square: Same, although seems obvious in retrospect.

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -66,6 +67,7 @@ public struct ListReorderGesture : Element
         isEnabled : Bool = true,
         actions : ReorderingActions,
         begins: Begins = .onTap,
+        reorderItemAccessibilityLabel: String? = nil,
```
meherkasam-square, line 70: Might be helpful to add docs for the parameters and indicate to the caller what a recommended localised reordering label might be.

meherkasam-square: nit: I noticed below that we're using a string format and plugging in the item name. What do you think about making this an enum instead where the caller can specify whether it's just the item name and they use our localised terminology or whether they themselves want to provide the terminology. For example, in case someone might want to call it `Rearrange Foo` instead of `Reorder Foo`.

BlueprintUILists/Sources/ListReorderGesture.swift
```diff
@@ -129,25 +130,89 @@ fileprivate extension ListReorderGesture
...
         let recognizer : ItemReordering.GestureRecognizer
+        private lazy var proxyElement = UIAccessibilityElement(accessibilityContainer: self)
+        private var minimumPressDuration: TimeInterval = 0.0 {
+            didSet {
+                updateGesturePressDuration()
+            }
+        }
+        
+        @objc private func updateGesturePressDuration() {
+            self.recognizer.minimumPressDuration = UIAccessibility.isVoiceOverRunning ? 0.0 : self.minimumPressDuration
```
meherkasam-square, line 144: Maybe not for this PR, but wondering how the behaviour would be with full keyboard access, Switch Control, and Voice Control.

meherkasam-square (approved):
Can you please add an entry to the changelog?

BlueprintUILists/Sources/ListReorderGesture.swift
RoyalPineapple: I wouldnt expect the gesture recognizer itself to work well with keyboard access. do the accessibility actions work with full keyboard access?

BlueprintUILists/Sources/ListReorderGesture.swift
RoyalPineapple: I'm going to push this entirely down to the caller, i think you're right that it needs more customizability but i think that too much magic is overcomplicated here.

PR #534, opened 2024-03-26
Add `backgroundView` to `ListProperties`

blakemcanally:
Adds `ListProperties.backgroundView` to enable developers to set a background on the underlying `UICollectionView`. This can be useful for more detailed layouts.

### Screenshots

Screenshot 2024-03-26 at 10 15 05 AM

### Checklist

### DIFF
--- a/ListableUI/Sources/ListProperties.swift
+++ b/ListableUI/Sources/ListProperties.swift
@@ -92,6 +92,9 @@ import UIKit
     /// The appearance to use with the list.
     public var appearance : Appearance
     
+    /// The view to use as the background of the list view.
+    public var backgroundView : UIView?
+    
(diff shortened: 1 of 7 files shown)


ListableUI/Sources/ListProperties.swift
kyleve, line 1: Sorry I'm just getting to this, but I'm not sure this is useful to expose – I believe we've solved the original case/need here, and generally in a Blueprint or SwiftUI world, we'd do this by composing layers together – eg you put a view _behind_ the list in an overlay, vs exposing this API directly.

blakemcanally: ...I've actually forgotten why I wanted this in the first place ¯\_(ツ)_/¯

PR #548, opened 2024-12-12
fix: Remove static view caching to improve memory management

robmaceachern:
- **Remove static caches from ListProperties**
- **Remove unused static cache vars**

The static `ReusableViewCache`s used in `ListProperties` would cause views to be retained beyond the lifetime of the owning list and if the views themselves referenced other objects (directly or through closures) then large parts of application state could be inadvertently "leaked" for the lifetime of the application.

This change instantiates new caches every time `PresentationState` is instantiated. There is a measuring performance impact since each reuse identifier will need to be recreated once per layout. On a long homogenous list that probably isn't a big deal, but for a heterogenous list it could be more of an issue (in the extreme case of a list composed of unique cells for every row this caching would provide no benefit). I haven't done any formal profiling on this.

Ideally we'd be able to associate the caches to the list views themselves so that their lifetimes were bound together. That would help avoid many of the memory issues we've seen while improving the cache effectiveness.

### Checklist

### DIFF
--- a/ListableUI/Sources/Layout/ListLayout/ListLayout+Layout.swift
+++ b/ListableUI/Sources/Layout/ListLayout/ListLayout+Layout.swift
@@ -11,9 +11,6 @@ import UIKit
 
 extension ListProperties {
     
-    private static let headerFooterMeasurementCache = ReusableViewCache()
-    private static let itemMeasurementCache = ReusableViewCache()
-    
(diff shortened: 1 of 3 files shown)


robmaceachern:
Closing. We decided to start with a manual cache clearing approach: https://github.com/square/Listable/pull/549

PR #552, opened 2025-01-08
Remove animation parameter, switch to using animated on UICollectionView directly. Add onDidEndDeceleration

cae7ffb:
Animating in a UIView animation block causes artifacts when scrolling. Remove animation blocks, and replace with simple animated parameter. Also introduce onDidEndDeceleration.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -606,8 +606,7 @@ public final class ListView : UIView
         animated: Bool = false
     ) -> Bool
     {
-        self.logHorizontalScrollToWarning()
-        
+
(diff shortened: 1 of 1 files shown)


ListableUI/Sources/ListActions.swift
```diff
@@ -94,8 +94,7 @@ public final class ListActions {
         public func scrollTo(
             item : AnyItem,
             position : ScrollPosition,
-            animation : ViewAnimation = .none,
```
kyleve, line 97: is `ViewAnimation` used anywhere still, or could we delete it?

kyleve (approved):

PR #557, opened 2025-02-04
Fix for crash when doing a programmatic scroll down while another external update is processed

kyleve:
This fixes an issue with multiple updates coming in close to each other and accidentally "undoing" the programmatic scroll down event that was previously enqueued. The fix is to instead of just referencing the first visible index path when processing further updates, to use the in-progress index path from the prior programmatic scroll event.

I've added a test that fails without the fix.

https://block.atlassian.net/browse/UI-8010

### Checklist

- [x] Unit tests

### DIFF
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -1057,6 +1057,7 @@ public final class ListView : UIView
         
         let indexPaths = self.collectionView.indexPathsForVisibleItems
         
+        // TODO: Pick biggest index path here? Does it matter?
         let indexPath = updateOverrideIndexPath ?? indexPaths.first
         
         let presentationStateTruncated = self.storage.presentationState.containsAllItems == false
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -1042,6 +1042,8 @@ public final class ListView : UIView
     // MARK: Internal - Updating Presentation State
     //
     
+    private var updateOverrideIndexPath : IndexPath? = nil
+    
     internal func updatePresentationState(
         for reason : PresentationState.UpdateReason,
         completion callerCompletion : @escaping (Bool) -> () = { _ in }
@@ -1055,7 +1057,7 @@ public final class ListView : UIView
         
         let indexPaths = self.collectionView.indexPathsForVisibleItems
         
-        let indexPath = indexPaths.first
+        let indexPath = updateOverrideIndexPath ?? indexPaths.first
         
         let presentationStateTruncated = self.storage.presentationState.containsAllItems == false
         
@@ -1090,7 +1092,13 @@ public final class ListView : UIView
             self.updatePresentationStateWith(firstVisibleIndexPath: indexPath, for: reason, completion: completion)
             
         case .programaticScrollDownTo(let scrollToIndexPath):
-            self.updatePresentationStateWith(firstVisibleIndexPath: scrollToIndexPath, for: reason, completion: completion)
+            
+            updateOverrideIndexPath = scrollToIndexPath
+            
+            self.updatePresentationStateWith(firstVisibleIndexPath: scrollToIndexPath, for: reason, completion: {
+                self.updateOverrideIndexPath = nil
+                completion($0)
+            })
         }
     }
         
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -1097,7 +1097,14 @@ public final class ListView : UIView
             updateOverrideIndexPath = scrollToIndexPath
             
             self.updatePresentationStateWith(firstVisibleIndexPath: scrollToIndexPath, for: reason, completion: {
-                self.updateOverrideIndexPath = nil
+                
+                /// Verify this is the same as inputted index path – if it's not, that means
+                /// _another_ `programaticScrollDownTo` has occurred and thus has
+                /// overridden this value, so we shouldn't clear it out.
+                if self.updateOverrideIndexPath == scrollToIndexPath {
+                    self.updateOverrideIndexPath = nil
+                }
+
                 completion($0)
             })
         }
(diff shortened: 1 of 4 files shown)


ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1081,7 +1083,13 @@ public final class ListView : UIView
             self.updatePresentationStateWith(firstVisibleIndexPath: indexPath, for: reason, completion: completion)
             
         case .programaticScrollDownTo(let scrollToIndexPath):
-            self.updatePresentationStateWith(firstVisibleIndexPath: scrollToIndexPath, for: reason, completion: completion)
+            
+            updateOverrideIndexPath = scrollToIndexPath
+            
+            self.updatePresentationStateWith(firstVisibleIndexPath: scrollToIndexPath, for: reason, completion: {
+                self.updateOverrideIndexPath = nil
```
kyleve, line 1090: TODO: Do we want to compare this to the `scrollToIndexPath`, and if it's changed, not touch it?

Podfile.lock
kyleve, line 1: Podfile was out of date

kyleve, line 1: This was out of date.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1081,7 +1084,13 @@ public final class ListView : UIView
             self.updatePresentationStateWith(firstVisibleIndexPath: indexPath, for: reason, completion: completion)
             
         case .programaticScrollDownTo(let scrollToIndexPath):
-            self.updatePresentationStateWith(firstVisibleIndexPath: scrollToIndexPath, for: reason, completion: completion)
+            
+            updateOverrideIndexPath = scrollToIndexPath
```
kyleve, line 1088: - [ ] Another thing to try...

```swift
self.performBatchUpdates(
    with: diff,
    animated: reason.animated,
    updateBackingData: updateBackingData,
    collectionViewUpdateCompletion: completion.finish,
    animationCompletion: callerCompletion
)
```
Move the `completion.finish` call into `animationCompletion` instead – as `completion.finish` is what finishes the async event queue event. But I think this would mean that _every_ update ends up waiting until the animation is done, which isn't desirable. Perhaps maybe we only do it for programmatic scroll down to?

kyleve: Hmm this didn't work – ah well!

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1046,7 +1048,8 @@ public final class ListView : UIView
         
         let indexPaths = self.collectionView.indexPathsForVisibleItems
```
robmaceachern, line 1049: Maybe a tangent: It looks like `indexPaths` isn't sorted. Are we expecting to use the lowest index path in downstream code? Based on the parameter name `firstVisibleIndexPath` it looks like it.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1081,7 +1084,20 @@ public final class ListView : UIView
...
+            
+            updateOverrideIndexPath = scrollToIndexPath
+            
+            self.updatePresentationStateWith(firstVisibleIndexPath: scrollToIndexPath, for: reason, completion: {
+                
+                /// Verify this is the same as inputted index path – if it's not, that means
+                /// _another_ `programaticScrollDownTo` has occurred and thus has
+                /// overridden this value, so we shouldn't clear it out.
+                if self.updateOverrideIndexPath == scrollToIndexPath {
+                    self.updateOverrideIndexPath = nil
```
robmaceachern, line 1096: I'm trying to think if there's a case where this breaks down (and don't get cleared out) but it _seems_ like it should work...

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1033,6 +1033,8 @@ public final class ListView : UIView
     // MARK: Internal - Updating Presentation State
     //
     
+    private var updateOverrideIndexPath : IndexPath? = nil
```
robmaceachern, line 1036: Probably worth adding some context around this property.

robmaceachern (approved):
It's hard for me to tell what downstream impacts there might be by passing an index path to `updatePresentationStateWith(firstVisibleIndexPath:...)` that isn't necessarily in the set of visible index paths. But this does seem to fix the crash we've seen.

Might be worth doing a test integration prior to landing this to see if the change breaks any existing register tests.

ListableUI/Sources/ListView/ListView.swift
kyleve: TIL!

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1046,7 +1048,8 @@ public final class ListView : UIView
         
         let indexPaths = self.collectionView.indexPathsForVisibleItems
         
-        let indexPath = indexPaths.first
+        // TODO: Pick biggest index path here? Does it matter?
```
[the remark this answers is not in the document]
kyleve: Fixed

ListableUI/Sources/ListView/ListView.swift
kyleve: Done

ListableUI/Sources/ListView/ListView.swift
kyleve: Yeah I think it's fine?

kyleve:
Integration: https://github.com/squareup/ios-register/pull/121918

kyleve:
Ugh need to move to a new github action, prior art here https://github.com/square/Blueprint/pull/523

ListableUI/Sources/ListView/ListView.swift
kyleve: Fixed!

kyleve:
> Integration: [squareup/ios-register#121918](https://github.com/squareup/ios-register/pull/121918)

Green! https://square-console.sqprod.co/bbr/production/ci_jobs/PR+ios+ZGTYUV+Bazel-Pipeline-PR/1

PR #558, opened 2025-02-24
Only animate scroll if UIView.areAnimationsEnabled

cae7ffb:
https://github.com/square/Listable/pull/552 changed how scrolling worked. The new method behaves differently when wrapped in `UIView.performWithoutAnimation` blocks, which impacts things like tests.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -537,7 +537,10 @@ public final class ListView : UIView
         guard let toIndexPath = self.storage.allContent.firstIndexPathForItem(with: item) else {
             return false
         }
-        
+
+        // If user is performing this in a `UIView.performWithoutAnimation` block, respect that and don't animate, regardless of what the animated parameter is.
+        let shouldAnimate = animated && UIView.areAnimationsEnabled
+
         return self.preparePresentationStateForScroll(to: toIndexPath) {
             
             /// `preparePresentationStateForScroll(to:)` is asynchronous in some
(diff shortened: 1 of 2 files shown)


ListableUI/Sources/ListView/ListView.swift
```diff
@@ -571,13 +574,13 @@ public final class ListView : UIView
...
                 )
-                self.performScroll(to: itemFrameAdjustedForStickyHeaders, scrollPosition: position, animated: animated)
+                self.performScroll(to: itemFrameAdjustedForStickyHeaders, scrollPosition: position, animated: shouldAnimate)
 
             } else {
                 self.collectionView.scrollToItem(
                     at: toIndexPath,
                     at: position.position.toUICollectionViewScrollPosition(for: self.collectionViewLayout.layout.direction),
-                    animated: animated
+                    animated: shouldAnimate
```
kyleve, line 583: Huh, weird this doesn't support / check `UIView.areAnimationsEnabled`, ah well

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -557,6 +557,9 @@ public final class ListView : UIView
                 return
             }
 
+            // If user is performing this in a `UIView.performWithoutAnimation` block, respect that and don't animate, regardless of what the animated parameter is.
+            let shouldAnimate = animated && UIView.areAnimationsEnabled
```
kyleve, line 561: FYI you may want to capture this outside of the `preparePresentationStateForScroll` call, since it can be async?

PR #560, opened 2025-03-18
Adding peek support to `PagedListLayout`

johnnewman-square:
- `PagedListLayout` now supports a peek value, which allows items to peek into view from the leading/top and trailing/bottom edges.
   - The peek can be disabled, uniform on both sides, or have a unique first item leading peek.
   - The layout's `PagingSize` now supports `inset(Peek)` and `fixed(CGFloat)` page sizes.
- Replacing the `isPagingEnabled` boolean with a `PagingStyle` enum. This gives layouts the ability to disable paging, leverage `native` UIScrollView paging (where items are full width), or use `custom` paging (where items can be any width).
- `ListScrollPositionInfo` now contains the percentage of visibility for each visible item. This allows clients to pick the most visible item for selection.

### Checklist

https://github.com/user-attachments/assets/ca81193a-0b3e-4ba9-a8b7-088677236f2d


johnnewman-square:
To ensure CI is working, this PR includes the actions/cache v4 changes from https://github.com/square/Listable/pull/561.

ListableUI/Sources/Behavior.swift
```diff
@@ -39,8 +39,8 @@ public struct Behavior : Equatable
     /// A Boolean value that determines whether the scroll view delays the handling of touch-down gestures.
     public var delaysContentTouches : Bool
     
-    /// Is paging enabled on the underlying scroll view.
-    public var isPagingEnabled : Bool
+    /// The paging style of underlying scroll view. When `.none`, no paging is performed.
+    public var pagingStyle : PagingStyle?
```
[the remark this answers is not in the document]
johnnewman-square: Agreed, thanks for the suggestion. That'll align better with other enums I see in Listable that explicitly define `none`.
- [ ] Add a `none` case to `PagingStyle` and make it required.

ListableUI/Sources/Internal/CGRect.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+extension CGRect {
+    
+    /// Returns the percentage from `0.0` to `1.0` that this rect overlaps `container`.
+    func percentageVisible(inside container: CGRect) -> CGFloat {
+        let overlap = intersection(container)
+        return (overlap.width * overlap.height) / (width * height)
```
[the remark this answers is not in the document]
johnnewman-square: Glad you called this out! I'll add safeguards for this and additional unit tests. I'd like to be extra cautious in case there is ever a zero-sized item.
- [ ] Divide by zero checks & tests

ListableUI/Sources/Layout/ListLayout/ListLayoutScrollViewProperties.swift
```diff
@@ -112,3 +112,16 @@ public enum ContentInsetAdjustmentBehavior : Equatable {
...
     }
 }
+
+
+/// The paging style of the scroll view.
+public enum PagingStyle {
+    
+    /// Applies native `UIScrollView` paging, where each page is the full width of the
+    /// scroll view's bounds.
+    case native
```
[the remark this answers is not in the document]
johnnewman-square: - [ ] Adjust `PagingStyle` name.

ListableUI/Sources/Behavior.swift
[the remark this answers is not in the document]
kyleve: +1, I'd prefer an explicit case personally

ListableUI/Sources/Internal/CGRect.swift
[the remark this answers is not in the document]
kyleve: +1 I'd recommend guarding against that just in case. I've found there's a bunch of very weird collection view corner cases that are similar

ListableUI/Sources/Layout/ListLayout/ListLayout.swift
```diff
@@ -507,25 +510,65 @@ extension AnyListLayout
...
+                horizontal: { velocity.x }
+            )
             
-            switch scrollDirection {
-            case .forward:
-                return edge >= offset
-            case .backward:
-                return edge <= offset
+            if mainAxisVelocity == 0 {
+                return items
```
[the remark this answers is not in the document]
kyleve: At least this is only in `itemToScrollToOnDidEndDragging`, so it will only be called on touch up for the cases this feature is enabled – and we did also sort previously above, too.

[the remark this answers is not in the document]
johnnewman-square: That's a fair point. My plan was to avoid refactoring the existing `.sorted` and `.first` logic because the next item in those cases may be offscreen due to the hardcoded width/height of `1000` in the query rect from `rectForFindingItemToScrollToOnDidEndDragging(after:velocity:)`.

For cases where the `pagingStyle` is `custom` (or peek), we should see some performance benefits if we use the `visibleContentFrame` rect as the query rect, since there should always be at least one item in the viewport. I'll mull this one over and work on an enhancement for the new paging style!

ListableUI/Sources/Layout/ListLayout/ListLayoutScrollViewProperties.swift
```diff
@@ -112,3 +112,16 @@ public enum ContentInsetAdjustmentBehavior : Equatable {
         }
     }
 }
+
+
+/// The paging style of the scroll view.
+public enum PagingStyle {
```
kyleve, line 118: I wonder if there's another name for this `PageScrollingBehavior` or something?

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
```diff
@@ -122,6 +159,61 @@ public struct PagedAppearance : ListLayoutAppearance
...
+        public var isEmpty: Bool {
+            value == 0 && firstItemLeadingValue == 0
+        }
+        
+        public init(value: CGFloat = 0, firstItemConfiguration: FirstItemConfiguration = .uniform) {
+            self.value = value
+            self.firstItemConfiguration = firstItemConfiguration
+        }
+        
+        /// This represents no peeking functionality.
```
kyleve, line 212: Nit: none instead of zero?

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
```diff
@@ -195,14 +279,27 @@ final class PagedListLayout : ListLayout
...
-            defaultWidth: bounds.width.clamp(pageSize.width)
-        )
         
-        var lastMaxY : CGFloat = 0
+        /// Apply the leading peek to the first item's position.
         
-        for item in content.all {
+        var lastMaxY : CGFloat = layoutAppearance.peek.firstItemLeadingValue
+        
+        for (index, item) in content.all.enumerated() {
```
kyleve, line 287: FYI we have a `forEachWithIndex` as well if you want it, but this is fine too

ListableUI/Sources/ListScrollPositionInfo.swift
```diff
@@ -22,7 +22,18 @@ public struct ListScrollPositionInfo : Equatable {
     //
     
     /// Which items within the list are currently visible.
-    public var visibleItems : Set<AnyIdentifier>
+    public var visibleItems : Set<VisibleItem>
+    
+    /// The item from `visibleItems` that has the highest percentage of visibility.
+    public var mostVisibleItem: VisibleItem? {
+        var mostVisibleItem: VisibleItem?
+        for visibleItem in visibleItems {
```
kyleve, line 30: Nit: Could this be a `reduce`?

kyleve (approved):
LGTM (I think), nice work! I want to think through the ListLayout changes a bit more, but they look right to me

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
johnnewman-square: - [ ] Replace `Peek.zero` with `Peek.none`

ListableUI/Sources/ListScrollPositionInfo.swift
johnnewman-square: Good call, that'll reduce the implementation a bit 😄 
- [ ] Use `reduce` and add a unit test for the `mostVisibleItem` API.

ListableUI/Sources/Layout/ListLayout/ListLayoutScrollViewProperties.swift
johnnewman-square: `PageScrollingBehavior` fits well since this enum is used for scroll offset calculations and for enabling native paging on the scroll view. Thanks for the suggestion.

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
johnnewman-square: Nice! I'll update this to use `forEachWithIndex` so that it matches Listable's styling.
- [ ] Use `forEachWithIndex`

Demo/Sources/Demos/Demo Screens/AutoScrollingViewController.swift
```diff
@@ -76,7 +76,7 @@ final class AutoScrollingViewController : UIViewController
                     shouldPerform: { info in
                         // Only scroll to the bottom if the bottom item is already visible.
                         if let identifier = lastItem {
-                            return info.visibleItems.contains(identifier)
+                            return info.visibleItems.map({ $0.identifier }).contains(identifier)
```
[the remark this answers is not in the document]
johnnewman-square: Updated to use keypaths, thanks for the suggestion 👏

ListableUI/Sources/Layout/ListLayout/ListLayout.swift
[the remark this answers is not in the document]
johnnewman-square: I pushed up a change that will use the `visibleContentFrame` for the query rect when the layout's scroll behavior is set to `peek`. This avoids iterating over offscreen elements and skips an item filtering pass before picking the offset.

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
```diff
@@ -122,6 +159,61 @@ public struct PagedAppearance : ListLayoutAppearance
...
+public extension PagedAppearance {
+    
+    /// This data model is used to apply an inset to each page, allowing items residing on the
+    /// edge of the collection view to "peek" into view.
+    struct Peek: Equatable {
+        
+        /// This is the main leading and trailing peek value.
+        public var value: CGFloat
+        
+        public var firstItemConfiguration: FirstItemConfiguration
```
nsillik, line 171: Should these be `let`s?

nsillik: actually, should these even be `public`? We don't seem to read them in the Market PR.

nsillik: Also, if `public` this needs a doc domment.

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
```diff
@@ -122,6 +159,61 @@ public struct PagedAppearance : ListLayoutAppearance
...
+        
+        /// This is the leading peek value before the first item.
+        var firstItemLeadingValue: CGFloat {
+            switch firstItemConfiguration {
+            case .uniform: value
+            case .customLeading(let customValue): customValue
+            }
+        }
+        
+        public enum FirstItemConfiguration: Equatable {
```
nsillik, line 181: Also needs a doc comment.

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
```diff
@@ -122,6 +159,61 @@ public struct PagedAppearance : ListLayoutAppearance
...
+        func totalValue(_ isFirstItem: Bool) -> CGFloat {
+            (isFirstItem ? firstItemLeadingValue : value) + value
+        }
+        
+        /// This is `true` if there are no peek values.
+        public var isEmpty: Bool {
+            value == 0 && firstItemLeadingValue == 0
+        }
+        
+        public init(value: CGFloat = 0, firstItemConfiguration: FirstItemConfiguration = .uniform) {
```
nsillik, line 207: Also needs a doc comment.

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
johnnewman-square: You're right, thanks for pointing this out. This is a holdover from an earlier Market implementation.
- [ ] Make constant and internal.

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
johnnewman-square: - [ ] Document `FirstItemConfiguration`.

ListableUI/Sources/Layout/Paged/PagedListLayout.swift
johnnewman-square: - [ ] Document the `Peek` initializer.

PR #562, opened 2025-04-01
bug: Disable cell content interaction while swipe actions are visible.

nsillik:
We don't want cell content to be interactable while the swipe actions are visisble. All touches should close the swipe area (and perform the associated action if a button was tapped), nothing else.

### Checklist

### DIFF
--- a/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -15,7 +15,7 @@ extension ItemCell {
     final class ContentContainerView : UIView, UIGestureRecognizerDelegate {
 
         let contentView : Content.ContentView
-        
+
         private var configurations: [Side: SwipeConfiguration] = [:]
         
         private var swipeAccessibilityCustomActions: [Side: [AccessibilitySwipeAction]] = [:] {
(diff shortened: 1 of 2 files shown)


ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -286,6 +286,14 @@ extension ItemCell {
         private func set(state: SwipeActionState, animated: Bool = false) {
             swipeState = state
 
+            // We don't want any actions on the content view while our swipe actions are open.
+            self.contentView.isUserInteractionEnabled = switch state {
```
tabendsquare, line 290: Just saying it out loud to make sure I'm following: this doesn't impact the tap to close swipe actions behavior because the tap gesture associated with `handleTap` is added directly to this view, not the contentView.

tabendsquare (approved):

ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
nsillik: exactly correct

robmaceachern (approved):
lgtm. thanks for reporting @tabendsquare 🙏🏽

PR #568, opened 2025-04-15
Fix reordering crash introduced in 16.0

kyleve:
In https://github.com/square/Listable/pull/560 we introduced a `percentageVisible` property on our list position info to support peek-based page layouts – this is usually fine, but due to event ordering, and namely how it takes an extra runloop for the collection view layout to update  after `collectionView(_ collectionView: UICollectionView, moveItemAt from: IndexPath, to: IndexPath)` is called, we end up in a state where the list's storage is inconsistent with its layout. This means that our call to `ListStateObserver.perform(self.view.stateObserver.onItemReordered, "Item Reordered" ...)` would crash / be wrong until the layout occurred. The fix is to place the `Item Reordered` observer callback at the end of the update queue, so the collection view has time to "settle" internally and become consistent.

[Relevant Slack Thread](https://square.slack.com/archives/CTJ4UNLHF/p1744750475867669)

https://block.atlassian.net/browse/UI-8607

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Fixed
 
+- Fixed a crash that would occur when a user would reorder / move items between sections.
+
 ### Added
 
 ### Removed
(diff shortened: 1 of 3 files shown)


johnnewman-square (approved):

johnnewman-square:
I've tested this using the demo app on iOS 16, 17, and 18 and haven't experienced any crashing when reordering between sections 👍

PR #572, opened 2025-05-02
Addressing an AutoScrollAction issue when using VerticalLayoutGravity.bottom

johnnewman-square:
- Adjusting the underlying CollectionView's `contentSize` before auto-scrolling. This will scroll after  the  `contentSize` logic  within CollectionView ([here](https://github.com/square/Listable/blob/main/ListableUI/Sources/ListView/ListView.swift#L1711)) has adjusted the offset.
- Ensuring the items provided to the `didPerform` closure are synced with the visible items after scrolling.

### Checklist

- [x] Unit tests.

### DIFF
--- a/ListableUI/Sources/AutoScrollAction.swift
+++ b/ListableUI/Sources/AutoScrollAction.swift
@@ -172,7 +172,6 @@ extension AutoScrollAction
         
         /// Called when the list performs the insertion.
         var didPerform : (ListScrollPositionInfo) -> () { get set }
-        
     }
     
     /// Values used to configure the `scrollToItem(onInsertOf:)` action.
--- a/ListableUI/Sources/AutoScrollAction.swift
+++ b/ListableUI/Sources/AutoScrollAction.swift
@@ -150,9 +150,33 @@ extension AutoScrollAction
         }
     }
     
+    /// This protocol allows `ListView` to treat the `OnInsertedItem` and `Pin` behaviors
+    /// in a similar fashion.
+    public protocol Behavior {
+        
+        /// The item in the list to scroll to.
+        var destination : ScrollDestination { get set }
+        
+        /// The desired scroll position.
+        var position : ScrollPosition { get set }
+        
+        /// If the change should be animated.
+        ///
+        /// ### Note
+        /// The action will only be animated if it is animated, **and** the list update itself is
+        /// animated. Otherwise, no animation occurs.
+        var animated : Bool { get set }
+        
+        /// An additional check you may provide to approve or reject the scroll action.
+        var shouldPerform : (ListScrollPositionInfo) -> Bool { get set }
+        
+        /// Called when the list performs the insertion.
+        var didPerform : (ListScrollPositionInfo) -> () { get set }
+        
+    }
     
     /// Values used to configure the `scrollToItem(onInsertOf:)` action.
-    public struct OnInsertedItem
+    public struct OnInsertedItem: AutoScrollAction.Behavior
     {
         /// The item in the list to scroll to when the `insertedIdentifier` is inserted.
         public var destination : ScrollDestination
@@ -160,43 +184,26 @@ extension AutoScrollAction
         /// The identifier of the item for which the `AutoScrollAction` should be performed.
         public var insertedIdentifier : AnyIdentifier
         
-        /// The desired scroll position,
         public var position : ScrollPosition
         
-        /// If the change should be animated.
-        ///
-        /// ### Note
-        /// The action will only be animated if it is animated, **and** the list update itself is
-        /// animated. Otherwise, no animation occurs.
         public var animated : Bool
         
-        /// An additional check you may provide to approve or reject the scroll action.
         public var shouldPerform : (ListScrollPositionInfo) -> Bool
         
-        /// Called when the list performs the insertion.
         public var didPerform : (ListScrollPositionInfo) -> ()
     }
 
     /// Values used to configure the `pin(to:)` action.
-    public struct Pin
+    public struct Pin: AutoScrollAction.Behavior
     {
-        /// The item in the list to scroll to.
         public var destination : ScrollDestination
 
-        /// The desired scroll position,
         public var position : ScrollPosition
         
-        /// If the change should be animated.
-        ///
-        /// ### Note
-        /// The action will only be animated if it is animated, **and** the list update itself is
-        /// animated. Otherwise, no animation occurs.
         public var animated : Bool
         
-        /// An additional check you may provide to approve or reject the scroll action.
         public var shouldPerform : (ListScrollPositionInfo) -> Bool
         
-        /// Called when the list performs the insertion.
         public var didPerform : (ListScrollPositionInfo) -> ()
     }
 }
--- a/ListableUI/Sources/AutoScrollAction.swift
+++ b/ListableUI/Sources/AutoScrollAction.swift
@@ -150,9 +150,9 @@ extension AutoScrollAction
         }
     }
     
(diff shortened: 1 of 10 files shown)


johnnewman-square:
Note: there are similar existing issues with `AutoScrollAction.pin(...)` sending stale visible items to its `didPerform` closure. I'll address that in a separate PR.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1282,13 +1282,35 @@ public final class ListView : UIView
...
+                        ///     `AutoScrollAction.scrollToItem(onInsertOf:)` behaviors
+                        ///   - the initial set of items in the list trigger the `scrollToItem(onInsertOf:)`
+                        ///   - the resulting scroll position isn't at the bottom of the list
+                        ///
+                        /// Without calling `layoutIfNeeded`, the above scenario will reset the scroll position
+                        /// to the bottom, discarding this scroll update. This is because the system will
+                        /// asynchronously update the underlying `contentSize` as part of the initial layout,
+                        /// moments after this method is executed. The list's `contentSize` is overridden to
+                        /// keep the offset anchored to the bottom when using `VerticalLayoutGravity.bottom`.
+                        collectionView.layoutIfNeeded()
```
kyleve, line 1299: I'm not 100% sure, so thinking out loud: Do we want a full `layoutIfNeeded`, or do we just want to trigger a `CollectionViewLayout` layout?

johnnewman-square: Thanks for the suggestion. A layout of the `CollectionViewLayout` feels like it would also work. I'm not sure of an approach that would allow us to layout the `CollectionViewLayout` synchronously since invalidating the layout will trigger a layout on the next update cycle. I'll mull this one over, but let me know if there's an approach like that in Listable. I'd be happy to spin up a new solution.

Another idea that doesn't require calling `layoutIfNeeded()` is to wrap this scroll logic in an async dispatch, which queues the scrolling to happen after the underlying `contentSize` is set & adjusted. With this approach, I do like the synchronous flow that `layoutIfNeeded()` provides, since it makes validating the results a bit more straightforward.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1282,13 +1282,35 @@ public final class ListView : UIView
...
-                        info.didPerform(self.scrollPositionInfo)
+                        /// Perform a layout after an animationless scroll so that `CollectionViewLayout`'s
+                        /// `prepare()` function will synchronously execute before calling `didPerform`. Otherwise,
+                        /// the list's `visibleContent` and the resulting `scrollPositionInfo.visibleItems` will
+                        /// be stale.
+                        collectionView.layoutIfNeeded()
+                        info.didPerform(scrollPositionInfo)
                     }
                 }
             }
```
kyleve, line 1316: Can't comment on it (it's below), and I'm not sure anyone even relies on this, but do we need anything in the `case .pin(let pin):` below?

ListableUI/Sources/ListView/ListView.swift
kyleve: > invalidating the layout will trigger a layout on the next update cycle

I wonder if doing a completely empty update to `collectionView.performBatchUpdates(...)` would also synchronously trigger a layout?

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1275,37 +1275,49 @@ public final class ListView : UIView
...
-            if pin.shouldPerform(self.scrollPositionInfo) {
-                /// Only animate the scroll if both the update **and** the scroll action are animated.
-                let animated = pin.animated && animated
-                
-                if let destination = pin.destination.destination(with: self.content) {
-                    guard self.scrollTo(item: destination, position: pin.position, animated: animated) else { return }
-                    if animated {
-                        stateObserver.onDidEndScrollingAnimation { state in
-                            pin.didPerform(self.scrollPositionInfo)
-                        }
```
johnnewman-square, line 1306: Using `self` within this `onDidEndScrollingAnimation { ... }` closure was causing a `ListView` memory leak. This is fixed now that the `Pin` and `OnInsertedItem` behaviors funnel through one function, using `state.positionInfo`.

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: This consolidation is complete! Now both `AutoScrollAction` behaviors use the same scrolling logic.

ListableUI/Sources/ListView/ListView.swift
kyleve: I'd go w/ the performBatchUpdates call personally, as `layoutIfNeeded` can have unintended side effects down the tree by forcing layouts, but ultimately up to you!

johnnewman-square: That sounds good! I've updated this PR to use `performBatchUpdates(nil)`.

ListableUI/Sources/ListView/ListView.swift
kyleve: Ahh good find

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1275,37 +1275,49 @@ public final class ListView : UIView
...
+                        ///     `AutoScrollAction` behaviors
+                        ///   - the initial set of items in the list trigger the autoscroll
+                        ///   - the resulting scroll position isn't at the bottom of the list
+                        ///
+                        /// Without calling `layoutIfNeeded`, the above scenario will reset the scroll position
+                        /// to the bottom, discarding this scroll update. This is because the system will
+                        /// asynchronously update the underlying `contentSize` as part of the initial layout,
+                        /// moments after this method is executed. The list's `contentSize` is overridden to
+                        /// keep the offset anchored to the bottom when using `VerticalLayoutGravity.bottom`.
+                        collectionView.performBatchUpdates(nil)
```
kyleve, line 1306: I think you may actually want to call [this one](https://github.com/square/Listable/blob/main/ListableUI/Sources/ListView/ListView%2BiOS16.4Workaround.swift#L137)? For iOS 16 which still has the bug that code fixes. You can pass an empty `CollectionViewChanges` IIRC

kyleve (approved):
One comment on which method to call, otherwise LGTM! Nice work

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: Thanks for the info. I've pushed up a change in https://github.com/square/Listable/pull/572/commits/f82f3d882aa32dd554f458bb6e3fd7c2f6ae6f7d to add a private `performEmptyBatchUpdates()` function that uses an empty `CollectionViewChanges`.

kyleve: Nice, thanks!

ListableUI/Sources/AutoScrollAction.swift
```diff
@@ -150,53 +150,59 @@ extension AutoScrollAction
         }
     }
     
+    /// This protocol allows `ListView` to treat the `OnInsertedItem` and `Pin` behaviors
+    /// in a similar fashion.
+    public protocol Behavior {
```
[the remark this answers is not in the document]
johnnewman-square: Thanks for the feedback and good point - this protocol exists mostly for configuring the AutoScrollAction. I may go with `AutoScrollAction.Configuration` instead of `AutoScrollAction.Behavior`. Does `Configuration` feel like it fits a bit better?

PR #575, opened 2025-05-13
Fix crash when using Headings rotor and we reach start of headings

meherkasam-square:
Fixed a crash that occurred when, while the cursor is on the first heading, we are on the Headings VoiceOver rotor and swipe up with one finger.

### Before

https://github.com/user-attachments/assets/bd252b04-c859-4da8-8c0a-492ecaa90416

### After

https://github.com/user-attachments/assets/0cb4d2d9-3db6-4ab6-ba64-9733439f57c3

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Fixed
 
+- Fixed a crash that occurred when, while the cursor is on the first heading, we are on the Headings VoiceOver rotor and swipe up with one finger.
+
 ### Added
 
 - Added support for `onDisplay` and `onEndDisplay` callbacks for `HeaderFooters`.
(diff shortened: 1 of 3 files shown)


kyleve:
@meherkasam-square Do we still want to land this?

meherkasam-square:
> @meherkasam-square Do we still want to land this?

Thanks for flagging. I will shortly run an integration PR on this to verify that it's working okay before marking the PR as ready for review.

meherkasam-square:
Tested on an integration PR and ran all the tests successfully. Manually verified against SPOS using VoiceOver that it's working as expected.

johnnewman-square (approved):

PR #576, opened 2025-05-16
Bottom gravity and autoscroll improvements

johnnewman-square:
When AutoScrollAction is engaged, these changes will temporarily ignore the bottom gravity overrides while scrolling to the desired item. This is a lighter alternative to the previous implementation, which asked the collection view to lay out its items before scrolling.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -1291,24 +1291,27 @@ public final class ListView : UIView
                 if let destination = info.destination.destination(with: self.content) {
                     
                     if behavior.verticalLayoutGravity == .bottom {
-                        /// Perform an update to adjust the `contentSize` of the collection view before
-                        /// scrolling. This avoids an issue where:
-                        ///   - the list is first appearing with `VerticalLayoutGravity.bottom` and
-                        ///     `AutoScrollAction` behaviors
-                        ///   - the initial set of items in the list trigger the autoscroll
-                        ///   - the resulting scroll position isn't at the bottom of the list
+                        /// Temporarily ignore the bottom gravity offest overrides before scrolling. This
+                        /// avoids an issue where:
+                        ///   - the list has `VerticalLayoutGravity.bottom` and `AutoScrollAction` behaviors
+                        ///   - the list has offscreen items that haven't been sized
+                        ///   - the `AutoScrollAction` has been triggered
+                        ///   - the resulting scroll position will adjust the collection view's `contentSize`
+                        ///     as items are dequeued and sized
                         ///
-                        /// Without calling `layoutIfNeeded`, the above scenario will reset the scroll position
-                        /// to the bottom, discarding this scroll update. This is because the system will
-                        /// asynchronously update the underlying `contentSize` as part of the initial layout,
-                        /// moments after this method is executed. The list's `contentSize` is overridden to
-                        /// keep the offset anchored to the bottom when using `VerticalLayoutGravity.bottom`.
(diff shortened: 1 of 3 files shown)


ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1291,24 +1291,27 @@ public final class ListView : UIView
...
                     
-                    guard self.scrollTo(item: destination, position: info.position, animated: animated) else { return }
+                    guard self.scrollTo(item: destination, position: info.position, animated: animated) else {
+                        collectionView.ignoreBottomGravityOffsetOverride = false
+                        return
+                    }
                     if animated {
-                        stateObserver.onDidEndScrollingAnimation { state in
+                        stateObserver.onDidEndScrollingAnimation { [weak self] state in
+                            self?.collectionView.ignoreBottomGravityOffsetOverride = false
```
kyleve, line 1314: Is there a risk of these somehow getting out of order and we result in this being stuck on true or something? I wonder if we could/should to the "counter" based approach instead, eg something like

```swift

var isIgnoringBottomGravityOffsetOverride : Bool {
   ignoringBottomGravityOffsetOverride == 0
}

func beginIgnoringBottomGravityOffsetOverride() { 
   self.ignoringBottomGravityOffsetOverride++
}
func endIgnoringBottomGravityOffsetOverride() { ... }
```

This of course has other risks where if you _miss_ a call to end, you get stuck in the wrong direction. Just thinking out loud here...

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1317,6 +1320,7 @@ public final class ListView : UIView
                         /// the list's `visibleContent` and the resulting `scrollPositionInfo.visibleItems` will
                         /// be stale.
                         performEmptyBatchUpdates()
```
kyleve, line 1322: We still need this one I guess?

kyleve (approved):

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: Yep! This call to `performEmptyBatchUpdates()` will keep the visible items in sync. We have some unit tests from the previous PR that test this particular flow. They'll fail if we remove this call.

kyleve: kk, thanks!

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: That's a good point.

One fallback solution could be to track the timestamp of when `ignoreBottomGravityOffsetOverride` is set to true. The `contentSize`, `frame`, and `contentInset` overrides could check if the collection view has been ignoring the gravity behavior for over a threshold of time, in which case they'd flip `ignoreBottomGravityOffsetOverride` back to false.

This last-ditch fallback would work with edge cases where `ignoreBottomGravityOffsetOverride` was set to true more times than false. I could see this happening if one of the `onDidEndScrollingAnimation` callbacks were somehow discarded.

johnnewman-square: I'm thinking that we could release these changes as-is and keep the timeout fallback on standby if we need to add a failsafe in the future. I do like that the timeout adds a little bit of extra safety, but it's difficult to pick a good scroll animation timeout value in this context. And I haven't run into a case where the `ignoreBottomGravityOffsetOverride` flag is being left on an unexpected value.

kyleve: SGTM!

PR #582, opened 2025-06-17
Adding a programmatic scroll completion handler

johnnewman-square:
This PR adds a completion handler to the `scrollTo(item:position:animated:)` API, turning it into `scrollTo(item:position:animated:completion:)`.

This also adds unit tests to assert the expected items are visible in the completion callback. A new demo allows you to use this API in a number of layout contexts, leveraging console logs when the scroll completion handler is executed:
Screenshot 2025-06-17 at 12 19 01 PM Screenshot 2025-06-17 at 12 19 08 PM

### Checklist


ListableUI/Sources/ListStateObserver.swift
```diff
@@ -94,6 +94,12 @@ public struct ListStateObserver {
     }
 
     private(set) var onDidEndScrollingAnimation: [OnDidEndScrollingAnimation] = []
+    
+    /// This internal variable is used to house the completion handler of scrolling APIs.
+    /// This handler avoids using the `onDidEndScrollingAnimation` handler array, which
+    /// isn't emptied in cases where the `ListView` is used in a standard UIView interface
+    /// instead of a Blueprint interface.
+    internal var programmaticScrollCompletion: OnDidEndScrollingAnimation?
```
kyleve, line 102: Could we stick this on the list view instead or something? Eg, I worry about this getting smashed over by something... Thinking about what if we added a `addScrollCompletion` to the ListView itself or something, since it has state, so multiple could be registered then cleared out.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,110 @@ public final class ListView : UIView
...
+    private enum ScrollCompletionReason {
+        case cannotScroll
+        case scrolled(animated: Bool)
+    }
+    
+    /// This function is used by programmatic scrolling APIs that provide a scroll
+    /// completion handler. This will execute the `completion` handler after scrolling
+    /// is finished, or it will execute immediately if scrolling is not possible or if
+    /// animations are disabled.
+    private func enqueueScrollHandler(reason: ScrollCompletionReason, completion: ((ListScrollPositionInfo) -> Void)?) {
```
kyleve, line 782: Seeing this, I think I prefer registering scroll completions w/ the collection view over the state observer, since in theory that's all external observations and can be replaced at any time.

ListableUI/Sources/ListStateObserver.swift
johnnewman-square: The callback handler has been moved out of `ListStateObserver` and into `ListView`. It was also made private to `ListView`.

With this change, I added unit tests to simulate rapid back-to-back calls to `scrollTo(...)` with a completion handler. When this happens, the previous animated scroll operation gets cancelled and its handler is executed before the next operation is started. Because of this setup, I didn't convert the handler into an array - I'm happy to make this change, but it shouldn't be possible for the handlers to stack up this way.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,110 @@ public final class ListView : UIView
...
+                stateObserver.programmaticScrollCompletion = { [weak self] _ in
+                    guard let self = self else { return }
+                    /// Sync the `scrollPositionInfo` before executing the handler.
+                    self.performEmptyBatchUpdates()
+                    completion(self.scrollPositionInfo)
+                }
+            } else {
+                /// Sync the `scrollPositionInfo` before executing the handler.
+                performEmptyBatchUpdates()
+                completion(scrollPositionInfo)
```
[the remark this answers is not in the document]
johnnewman-square: Circling back: this has been updated to execute in an async closure.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,129 @@ public final class ListView : UIView
...
+    private enum ScrollCompletionReason {
+        case cannotScroll
+        case scrolled(animated: Bool)
+    }
+    
+    /// This function is used by programmatic scrolling APIs that provide a scroll
+    /// completion handler. This will execute the `completion` handler after scrolling
+    /// is finished, or it will execute immediately if scrolling is not possible or if
+    /// animations are disabled.
+    private func enqueueScrollHandler(reason: ScrollCompletionReason, completion: ScrollCompletion?) {
```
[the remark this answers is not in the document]
johnnewman-square: > I'd nit this is a little of a misnomer – enqueue kinda implies you can have arbitrarily many of these pending, when it's really just setting one.

Good point! This function has been renamed `handleScrollCompletion(reason:completion:)`.

> Is there any situation where that'd be a problem?

There are some new unit tests ([here](https://github.com/square/Listable/pull/582/files#diff-d4181dc057ced1cdc4d34137ca5e8eb4600b687ea0baad3ad37f80a339f9e974R1052)) for cases where a client may rapidly call `scrollTo(...)` which programmatically interrupts the scrolling animation. Those tests help ensure that each call's completion handler gets executed and not discarded. Similarly, if a user manually interrupts an active scrolling animation by touching down in the scroll view or tapping the status bar, those actions execute the completion handler when the scroll gets interrupted and then nullify the handler.

Those cases cover us from flows where the handlers could stack up, but I'm happy to make adjustments if you see an issue.

johnnewman-square:
I've pushed up a change to add the completion handler to the equivalent `ListActions` scrolling APIs so that they're accessible in both `ListView` and `ListActions`.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,129 @@ public final class ListView : UIView
...
+                    self.performEmptyBatchUpdates()
+                    completion(ListStateObserver.DidEndScrollingAnimation(positionInfo: self.scrollPositionInfo))
+                }
+            }
+        }
+    }
+    
+    /// This is used to house the completion handler of scrolling APIs. This is kept
+    /// private and separate from `ListStateObserver` and its handlers.
+    private var scrollCompletionHandler: ScrollCompletion?
```
kyleve, line 808: I think we want this to be an array we fill up then empty on run, right? Eg, what if someone enqueues two scroll events in quick succession, won't the first one get silently dropped? I think there's also some subtle behavior to consider here where really, only the last completion handler in that list will have fully completed its scroll event.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,129 @@ public final class ListView : UIView
...
+    
+    /// This is called by the `ListView.Delegate` and is used to notify the
+    /// `scrollCompletionHandler` that scrolling finished. This does nothing if there is
+    /// no handler set.
+    internal func didEndScrolling() {
+        if let handler = scrollCompletionHandler {
+            /// Sync the `scrollPositionInfo` before executing the handler.
+            performEmptyBatchUpdates()
```
kyleve, line 816: Fwiw I _think_ this is safe, but it is introducing a batch update call on the end of each scroll event, which could probably somehow change _some_ behavior. Worth doing a POS integration on this branch to ensure nothing weird breaks.

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: Thanks for the feedback! I haven't reproduced a case where the handlers can queue, but I'll do some more investigation - I'd like to capture that flow in a unit test if possible.

- [ ] Investigate handler queueing for a case that discards the original handler.

Here's what I've tested:
- If a user manually interrupts a `scrollTo(...)` animation by touching down in the scroll view, the completion handler will execute immediately, reporting the currently visible items.
- For programmatic interrupts, there are new unit tests in [`test_rapid_scroll_to_item_completion_calls()`](https://github.com/square/Listable/pull/582/files#diff-d4181dc057ced1cdc4d34137ca5e8eb4600b687ea0baad3ad37f80a339f9e974R1052) that capture the behavior of calls in quick succession:
   - Rapidly calling `scrollTo(...)` without an animation executes all handlers, reporting the final visible items.
   - When the target item is off screen, rapidly calling `scrollTo(...)` with an animation will execute all handlers, but only the handler from the most recent call to `scrollTo(...)` will report the item as visible.
   - When the target item is on screen, rapidly calling `scrollTo(...)` with an animation will execute all handlers and each handler will report that the item as visible.

johnnewman-square: I've pushed up an update to switch to a queue of scroll completion handlers. This also includes a test to simulate a handful of them being queued and executing at once.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,129 @@ public final class ListView : UIView
...
+    /// This is used to house the completion handler of scrolling APIs. This is kept
+    /// private and separate from `ListStateObserver` and its handlers.
+    private var scrollCompletionHandler: ScrollCompletion?
+    
+    /// This is called by the `ListView.Delegate` and is used to notify the
+    /// `scrollCompletionHandler` that scrolling finished. This does nothing if there is
+    /// no handler set.
+    internal func didEndScrolling() {
+        if let handler = scrollCompletionHandler {
+            /// Sync the `scrollPositionInfo` before executing the handler.
```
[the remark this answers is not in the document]
johnnewman-square: Docs have been updated!

kyleve (approved):

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,129 @@ public final class ListView : UIView
...
+    /// and will not execute its `scrollViewDidEndScrollingAnimation(_:)` delegate.
+    func willScroll(
+        for scrollPosition: UICollectionView.ScrollPosition,
+        itemFrame: CGRect,
+        viewport: CGRect,
+        contentSize: CGSize
+    ) -> Bool {
+        func roundRect(_ rect: CGRect) -> CGRect {
+            CGRect(
+                x: round(rect.minX),
```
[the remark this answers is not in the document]
johnnewman-square: This is a great point. I've refactored the `willScroll(...)` function to simplify how it measures the distance to scroll and whether there's room to change the offset. There's also a new unit test for testing a fractional content offset.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,123 @@ public final class ListView : UIView
...
+    
+    /// This is used to house the completion handlers of scrolling APIs. This is kept
+    /// internal and separate from `ListStateObserver` and its handlers.
+    internal var scrollCompletionHandlers: [ScrollCompletion] = []
+    
+    /// This is called by the `ListView.Delegate` and is used to notify the
+    /// `scrollCompletionHandler` that scrolling finished. This does nothing if there is
+    /// no handler set.
+    internal func didEndScrolling() {
+        if scrollCompletionHandlers.count > 0 {
```
robmaceachern, line 814: Nit: Perhaps a `guard scrollCompletionHandlers.isEmpty == false else { return }`? Avoids a level of indentation for the main code.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,123 @@ public final class ListView : UIView
...
+            scrollCompletionHandlers.removeAll()
+            
+            /// Sync the `scrollPositionInfo` before executing the handlers.
+            ///
+            /// Calling `performEmptyBatchUpdates()` will synchronously call the
+            /// `CollectionViewLayout`'s `prepare()` function. That will update the
+            /// list's `visibleContent`, which `scrollPositionInfo` uses to accurately
+            /// find the visible items.
+            ///
+            /// ListViewTests has unit tests to assert that the handler's items are correct.
```
robmaceachern, line 825: Nit: I think the convention is to use `//` rather than `///` for these implementation note comments. Not sure if it really matters but probably good to match existing style.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,123 @@ public final class ListView : UIView
...
+    /// This function will determine if a call to `collectionView.scrollToItem(...)`
+    /// will result in an adjusted content offset. This is necessary because when the
+    /// item is already at the expected position, `UICollectionView` will not scroll
+    /// and will not execute its `scrollViewDidEndScrollingAnimation(_:)` delegate.
+    func willScroll(
+        for scrollPosition: UICollectionView.ScrollPosition,
+        itemFrame: CGRect,
+        viewport: CGRect,
+        contentSize: CGSize
+    ) -> Bool {
```
robmaceachern, line 843: I'm just thinking through what it would mean if the scroll view behavior ever changed slightly and introduced a mismatch:
- We return `false` but the scroll view _does_ actually scroll
  - Completion will fire early
- We return `true` but the scroll view _doesn't_ actually scroll
  - Completion won't fire at all (right?)

Do you figure the risk of this getting out of sync is pretty low?

robmaceachern (approved):

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,123 @@ public final class ListView : UIView
...
+            DispatchQueue.main.async {
+                completion(ListStateObserver.DidEndScrollingAnimation(positionInfo: self.scrollPositionInfo))
+            }
+        case .scrolled(let animated):
+            if animated {
+                scrollCompletionHandlers.append(completion)
+            } else {
+                // Dispatch so that scrolling without an animation executes the closure
+                // on the next runloop execution, similar to scrolling with an animation.
+                DispatchQueue.main.async {
```
g-mark, line 797: Can we ever have a non-animated `handleScrollCompletion` when `scrollCompletionHandlers` isn't empty? I'm not sure it would really matter, but curious if this came up at all.

g-mark (approved):
LGTM!

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -740,6 +766,123 @@ public final class ListView : UIView
...
+    /// internal and separate from `ListStateObserver` and its handlers.
+    internal var scrollCompletionHandlers: [ScrollCompletion] = []
+    
+    /// This is called by the `ListView.Delegate` and is used to notify the
+    /// `scrollCompletionHandler` that scrolling finished. This does nothing if there is
+    /// no handler set.
+    internal func didEndScrolling() {
+        if scrollCompletionHandlers.count > 0 {
+            let handlers = scrollCompletionHandlers
+            scrollCompletionHandlers.removeAll()
```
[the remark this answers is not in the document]
johnnewman-square: I decided to proactively remove them before executing the handlers just in case one of the handlers had a synchronous side effect of executing another call to `scrollTo(...)`. I think this is unlikely, but I didn't want to accidentally erase a freshly queued handler.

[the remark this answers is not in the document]
robmaceachern: Makes sense. Might be worth adding that context in case this code is reworked in the future.

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: This has been switch to a `guard`.

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: The comments have been fixed.

PR #585, opened 2025-06-30
Adding a completion handler to the `scrollToSection` API.

johnnewman-square:
This builds on the changes in https://github.com/square/Listable/pull/582.

### Checklist

- [ ] Ensure any public-facing changes are reflected in the [changelog](https://github.com/square/Listable/blob/main/CHANGELOG.md). Include them in the `Main` section.


kyleve (approved):
Nice!

ListableUI/Tests/ListView/ListViewTests.swift
```diff
@@ -1049,6 +1049,112 @@ class ListViewTests: XCTestCase
...
+                            position: scrollPosition,
+                            ifAlreadyVisible: visibilityRule
+                        ),
+                        animated: animated,
+                        completion: { changes in
+                            positionInfo = changes.positionInfo
+                            scrollExpectation.fulfill()
+                        }
+                    )
+                    wait(for: [scrollExpectation], timeout: 0.5)
```
g-mark, line 1151: This is fine as is, but I wonder if we could either disable animations on the view (if it doesn't interfere with the scrolling behavior) or use something like `window.layer.speed = 4` to speed up the testing?

g-mark (approved):
Nice!

ListableUI/Tests/ListView/ListViewTests.swift
johnnewman-square: Thanks for the suggestion. I've set the window speed to 4 for these tests. That may reduce the test duration a bit in CI. (`UIView.setAnimationsEnabled(false)` causes unit tests to break, so I'll stick with the window speed for now.)

PR #588, opened 2025-07-02
Reset scroll position if list identifier changes

scelis:
If a List’s identifier changes, we should automatically scroll to the top because maintaining scroll position from one list to another doesn’t make sense if the identifier (and thus list itself) completely changes.

Reseting the scroll position to the top in this scenario should be done without animation.

https://block.atlassian.net/browse/UI-8965

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


kyleve (approved):

johnnewman-square (approved):
This looks good!

PR #594, opened 2025-07-29
Bumping versions to 16.4.0.

robmaceachern:
16.4.0 release.

[Changelog](https://github.com/square/Listable/blob/main/CHANGELOG.md)

### Checklist

- [ ] Post merge: run the following command to tag the release:
```
./Scripts/tag_release.sh 16.4.0 <merge commit SHA>
```


johnnewman-square (approved):
This looks good!

I think we'll want to add the new version to the bottom of the changelog and update the link to main as a final step.

robmaceachern:
> I think we'll want to add the new version to the bottom of the changelog and update the link to main as a final step.

Oops! dabb758d Thanks!

And it turns out that we've been doing these wrong since 15.0.1 so I fixed that too: 73e2e3ae

PR #595, opened 2025-07-30
fix: Ensure we're referencing up-to-date swipe state when configuring interaction enabled

robmaceachern:
For https://block.atlassian.net/browse/UI-9091

This resolves an issue where a cell would not respond to taps after the "full swipe" swipe action trigger.

The issue was that we were referencing stale swipe state when configuring `isUserInteractionEnabled`. It was quite unintuitive when debugging this, but the line `swipeState = state` immediately resulted in those variables holding different values while processing the swipe action 🙃. We would incorrectly fall into the `.willPerformFirstActionAutomatically` case, making the content view unresponsive, despite the `swipeState` reporting as `.closed`.

The SwipeActionsViewController has also been updated to demonstrate this is now working as expected.

### Checklist

- [ ] Ensure any public-facing changes are reflected in the [changelog](https://github.com/square/Listable/blob/main/CHANGELOG.md). Include them in the `Main` section.

### DIFF
--- a/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
+++ b/ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
@@ -286,8 +286,11 @@ extension ItemCell {
         private func set(state: SwipeActionState, animated: Bool = false) {
             swipeState = state
 
+            // Warning! : `swipeState` may be mutated to a new value via it's property
+            // observers and may not equal `state` at this point!
+
             // We don't want any actions on the content view while our swipe actions are open.
-            self.contentView.isUserInteractionEnabled = switch state {
+            self.contentView.isUserInteractionEnabled = switch swipeState {
             case .closed:
                 true
             case .expandActions, .open, .swiping, .willPerformFirstActionAutomatically:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


ListableUI/Sources/Internal/ItemCell.ContentViewContainer.swift
```diff
@@ -286,8 +286,11 @@ extension ItemCell {
         private func set(state: SwipeActionState, animated: Bool = false) {
             swipeState = state
 
+            // Warning! : `swipeState` may be mutated to a new value via it's property
```
kyleve, line 289: What's an example of this happening?

robmaceachern: In the original case it was: `willPerformFirstActionAutomatically(ListableUI.SwipeActionsView.Side.right)` becoming `closed`.

We get into this kind of re-entrant `set(state:)` situation if the swipe action completion is invoked synchronously.

image

I'm not sure if there's a better/safer way to restructure this stuff.

kyleve: Ahh interesting – synchronous completion handlers strike again... Got it!

There's probably some improvement, but this also seems like a conceptually fine/correct fix too!

kyleve (approved):

johnnewman-square (approved):
Nice find!

PR #615, opened 2026-05-21
Add custom item offset scrolling

aaalaniz:
Adds a Listable-owned API for custom vertical item positioning from both imperative `ListActions` and declarative `AutoScrollAction`.

### What's changed

- Adds `ListItemScrollPosition.standard(_:)` and `.verticalContentOffsetAdjustment(_:)`.
- Adds `AutoScrollAction.scrollTo(... itemPosition:)` and `pin(... itemPosition:)`, with existing `position:` overloads preserved.
- Adds `AutoScrollAction.ScrollInterruptionPolicy`:
  - `.performImmediately`
  - `.deferDuringUserScrolling`
  - `.skipDuringUserScrolling`
- Keeps custom adjustments vertical-only; no horizontal offset API is added.
- Adds a generic demo: `Custom Auto Scrolling (Footer-Aware Pin)`.
- Adds tests for custom declarative auto-scroll, callback behavior, deferral, and skip policy.

### Example

```swift
list.autoScrollAction = .pin(
    .item(targetIdentifier),
    itemPosition: .verticalContentOffsetAdjustment { info in
        max(0.0, info.itemFrame.maxY - info.visibleContentFrame.maxY)
    },
    scrollInterruptionPolicy: .deferDuringUserScrolling
)
```

Use `.skipDuringUserScrolling` when an auto-scroll request should be dropped instead of retried after user scrolling ends.

### Visuals

https://github.com/user-attachments/assets/177a8ffb-ae96-4860-938a-af93d5655a75

### Test Plan

- [x] `mise exec -- xcodebuild test -workspace Development/ListableDevelopment.xcworkspace -scheme ListableDevelopment-Workspace -destination 'id=52DC49DC-0376-4B43-99BE-BC7A016A51B3' -derivedDataPath /tmp/listable-dd-development-tests -only-testing:ListableTests`
- [x] `git diff --check`
- [x] Built and launched `Custom Auto Scrolling (Footer-Aware Pin)` demo on iOS Simulator

### Checklist

### DIFF
--- a/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
+++ b/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
@@ -21,6 +21,7 @@ final class CustomAutoScrollingViewController : UIViewController
     private var selectedRow = 24
     private var expandedRows = Set<Int>()
     private var hasPerformedInitialLayoutUpdate = false
+    private var animateAutoScroll = false
 
     override func loadView()
     {
@@ -53,6 +54,12 @@ final class CustomAutoScrollingViewController : UIViewController
         super.viewDidLoad()
 
         self.title = "Custom Auto Scrolling"
+        self.navigationItem.rightBarButtonItem = UIBarButtonItem(
+            title: "Toggle Animations",
+            style: .plain,
+            target: self,
+            action: #selector(toggleAnimations)
+        )
         self.updateList()
     }
 
@@ -142,6 +149,12 @@ final class CustomAutoScrollingViewController : UIViewController
         self.updateList()
     }
 
+    @objc private func toggleAnimations()
+    {
+        self.animateAutoScroll.toggle()
+        print("Auto-scroll animations are \(self.animateAutoScroll ? "on" : "off").")
+    }
+
     private func updateList()
     {
         self.footerTitle.text = "Target row \(self.selectedRow + 1) stays above the fixed footer"
@@ -160,7 +173,7 @@ final class CustomAutoScrollingViewController : UIViewController
                 itemPosition: .verticalContentOffsetAdjustment { [weak self] info in
                     self?.footerAwareScrollDelta(for: info) ?? 0.0
                 },
-                animated: false,
+                animated: self.animateAutoScroll,
                 scrollInterruptionPolicy: .deferDuringUserScrolling,
                 shouldPerform: { _ in true }
             )
--- a/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
+++ b/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
@@ -17,6 +17,14 @@ final class CustomAutoScrollingViewController : UIViewController
     private let list = ListView()
     private let footer = UIView()
     private let footerTitle = UILabel()
+    private lazy var animationsButton : UIBarButtonItem = {
+        UIBarButtonItem(
+            title: self.animationsButtonTitle,
+            style: .plain,
+            target: self,
+            action: #selector(toggleAnimations)
+        )
+    }()
 
     private var selectedRow = 24
     private var expandedRows = Set<Int>()
@@ -55,12 +63,7 @@ final class CustomAutoScrollingViewController : UIViewController
         super.viewDidLoad()
 
         self.title = "Custom Auto Scrolling"
-        self.navigationItem.rightBarButtonItem = UIBarButtonItem(
-            title: "Toggle Animations",
-            style: .plain,
-            target: self,
-            action: #selector(toggleAnimations)
-        )
+        self.navigationItem.rightBarButtonItem = self.animationsButton
         self.updateList()
     }
 
@@ -158,7 +161,13 @@ final class CustomAutoScrollingViewController : UIViewController
     @objc private func toggleAnimations()
     {
         self.animateAutoScroll.toggle()
+        self.updateAnimationsButtonTitle()
         print("Auto-scroll animations are \(self.animateAutoScroll ? "on" : "off").")
+    }
+
+    private func updateAnimationsButtonTitle()
+    {
+        self.animationsButton.title = self.animationsButtonTitle
     }
 
     private func updateList()
@@ -222,6 +231,10 @@ final class CustomAutoScrollingViewController : UIViewController
 
     private var scrollIndicatorBottomInset : CGFloat {
         max(0.0, self.footer.bounds.height - self.view.safeAreaInsets.bottom)
+    }
+
+    private var animationsButtonTitle : String {
+        "Animations: \(self.animateAutoScroll ? "On" : "Off")"
     }
 
     private static let rowCount = 50
--- a/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
+++ b/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
@@ -68,6 +68,13 @@ final class CustomAutoScrollingViewController : UIViewController
         self.updateList()
     }
 
+    override func viewSafeAreaInsetsDidChange()
+    {
+        super.viewSafeAreaInsetsDidChange()
+
+        self.updateList()
+    }
+
     private func configureFooter()
     {
         self.footer.backgroundColor = .systemBackground
@@ -146,7 +153,7 @@ final class CustomAutoScrollingViewController : UIViewController
             list.appearance = .demoAppearance
             list.layout = .demoLayout
             list.animation = .fast
-            list.scrollIndicatorInsets.bottom = 112.0
+            list.scrollIndicatorInsets.bottom = self.scrollIndicatorBottomInset
 
             list.autoScrollAction = .pin(
                 .item(targetIdentifier),
@@ -192,6 +199,10 @@ final class CustomAutoScrollingViewController : UIViewController
         }
 
         return 0.0
+    }
+
+    private var scrollIndicatorBottomInset : CGFloat {
+        max(0.0, self.footer.bounds.height - self.view.safeAreaInsets.bottom)
     }
 
     private static let rowCount = 50
--- a/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
+++ b/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
@@ -22,6 +22,7 @@ final class CustomAutoScrollingViewController : UIViewController
     private var expandedRows = Set<Int>()
     private var hasPerformedInitialLayoutUpdate = false
     private var animateAutoScroll = false
+    private var isSuppressingAutoScrollAnimation = false
 
     override func loadView()
     {
@@ -72,7 +73,12 @@ final class CustomAutoScrollingViewController : UIViewController
         }
 
         self.hasPerformedInitialLayoutUpdate = true
-        self.updateList()
+        self.isSuppressingAutoScrollAnimation = true
+        defer { self.isSuppressingAutoScrollAnimation = false }
+
+        UIView.performWithoutAnimation {
+            self.updateList()
+        }
     }
 
     override func viewSafeAreaInsetsDidChange()
@@ -173,7 +179,7 @@ final class CustomAutoScrollingViewController : UIViewController
                 itemPosition: .verticalContentOffsetAdjustment { [weak self] info in
                     self?.footerAwareScrollDelta(for: info) ?? 0.0
                 },
-                animated: self.animateAutoScroll,
+                animated: self.animateAutoScroll && self.isSuppressingAutoScrollAnimation == false,
                 scrollInterruptionPolicy: .deferDuringUserScrolling,
                 shouldPerform: { _ in true }
             )
--- a/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
+++ b/Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
@@ -1,0 +1,254 @@
+//
+//  CustomAutoScrollingViewController.swift
+//  Demo
+//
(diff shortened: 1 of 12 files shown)


aaalaniz:
🤖 @codex review

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+        let selectedRow = self.selectedRow
+        let targetIdentifier = FooterAwarePinnedItem.identifier(with: selectedRow)
+
+        self.list.configure { list in
+            list.appearance = .demoAppearance
+            list.layout = .demoLayout
+            list.animation = .fast
+            list.scrollIndicatorInsets.bottom = 112.0
+
+            list.autoScrollAction = .pin(
```
aaalaniz, line 186: The new API in action

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -635,6 +661,60 @@ public final class ListView : UIView
...
+    /// Scrolls to a custom vertical offset for the item with the provided identifier.
+    /// The adjustment receives the item's frame and visible content frame,
+    /// then returns the vertical delta to apply.
+    /// If there is more than one item with the same identifier, the list scrolls to the first.
+    /// If the item is contained in the list, true is returned. If it is not, false is returned.
+    ///
+    @discardableResult
+    public func scrollTo(
+        item : AnyIdentifier,
+        contentOffsetAdjustment : @escaping ListItemScrollPositionAdjustment,
```
aaalaniz, line 674: I think it's okay that this is marked as escaping but making a note for myself

ListableUI/Tests/ListView/ListViewTests.swift
```diff
@@ -846,6 +846,354 @@ class ListViewTests: XCTestCase
...
+                ))
+                waitFor { vc.list.updateQueue.isEmpty }
+
+                shouldPerform = true
+                vc.list.delegate.scrollViewDidEndDragging(vc.list.collectionView, willDecelerate: false)
+                waitFor { didPerform.count == 1 }
+            }
+        }
+
+        self.testcase("skips while dragging") {
```
aaalaniz, line 1162: Add coverage when user specifies to skip scroll during scroll

ListableUI/Sources/ListView/ListView.Delegate.swift
johnnewman-square: I received this feedback from Codex. I think we could potentially make a few adjustments here and add a unit test to capture this specific case:
> In ListView.Delegate.swift, onDidEndScrollingAnimation now gets actions, so a client can start another scroll from inside that callback. But self.view.didEndScrolling() is called after the observer callback. If the callback starts an animated scroll with a completion, ListView.swift appends that new completion, and then the current “old scroll ended” event drains it immediately. So the completion can fire with stale position info before the newly requested scroll finishes.
> 
> The likely fix is to snapshot/drain existing scroll completions before invoking onDidEndScrollingAnimation, or otherwise make didEndScrolling() only process handlers that existed before this delegate callback began.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+        self.footerTitle.text = "Target row \(self.selectedRow + 1) stays above the fixed footer"
+
+        let selectedRow = self.selectedRow
+        let targetIdentifier = FooterAwarePinnedItem.identifier(with: selectedRow)
+
+        self.list.configure { list in
+            list.appearance = .demoAppearance
+            list.layout = .demoLayout
+            list.animation = .fast
+            list.scrollIndicatorInsets.bottom = 112.0
```
johnnewman-square: We could replace the hardcoded inset with an example of a calculated inset that gets the scroll indicator closer to the top edge of the footer. Though, we'd need to call `updateList` in `viewSafeAreaInsetsDidChange()` for good measure. Up to you!
```suggestion
            list.scrollIndicatorInsets.bottom = footer.bounds.height - view.safeAreaInsets.bottom
```

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+
+            list.autoScrollAction = .pin(
+                .item(targetIdentifier),
+                itemPosition: .verticalContentOffsetAdjustment { [weak self] info in
+                    self?.footerAwareScrollDelta(for: info) ?? 0.0
+                },
+                animated: false,
+                scrollInterruptionPolicy: .deferDuringUserScrolling,
+                shouldPerform: { _ in true }
+            )
```
johnnewman-square, line 194: I think we should explore the initial animations when the view is presented. When opening the demo, the list animates to the selected row when `viewDidLayoutSubviews()` calls `updateList()`. It would be a great enhancement if we could find a way around this issue in the demo example. Also curious if this is an issue in the production use case.

Here's a recording with slow animations, taken from an iOS 17.5 simulator.

https://github.com/user-attachments/assets/6d10b5b6-5bc8-4c22-b8cc-1a481e15d3df

While investigating, I flipped animations to enabled by default in `AutoScrollingViewController3` but opening that demo's controller didn't perform the same initial scroll there. I don't think this is an issue in the other demos that showcase `pin` and `scrollTo` because they don't have the `viewDidLayoutSubviews` pass.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+            list.layout = .demoLayout
+            list.animation = .fast
+            list.scrollIndicatorInsets.bottom = 112.0
+
+            list.autoScrollAction = .pin(
+                .item(targetIdentifier),
+                itemPosition: .verticalContentOffsetAdjustment { [weak self] info in
+                    self?.footerAwareScrollDelta(for: info) ?? 0.0
+                },
+                animated: false,
```
johnnewman-square: A button in the navigation bar to toggle the animated flag would be a neat enhancement here, similar to some of the other autoscrolling demos.

johnnewman-square (commented):
This is a neat enhancement. I left just a few notes that I think we could potentially explore.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
aaalaniz: Good suggestion. This is more aligned with the target use case too 👍🏼

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
aaalaniz: Makes sense, I'll add.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
aaalaniz: I will investigate 👍🏼

ListableUI/Sources/ListView/ListView.Delegate.swift
aaalaniz: Good catch, will adjust.

aaalaniz: 🤖 [Codex] Addressed in bd63c5a2. The delegate now drains the completion handlers that existed before `onDidEndScrollingAnimation`, runs the observer callback with actions, and then executes only that drained batch. I also added a regression test covering a new animated scroll started from the callback so its completion waits for the next scroll end.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
aaalaniz: 🤖 [Codex] Addressed in c27fe330. The demo now calculates the bottom scroll indicator inset from the footer height minus the safe-area bottom inset, clamps it to zero, and re-runs `updateList()` from `viewSafeAreaInsetsDidChange()`.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
aaalaniz: 🤖 [Codex] Addressed in 67ade56a. The demo now suppresses animation only for the first layout-driven `updateList()` pass when the screen is presented. Later row changes still honor the animation toggle.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
aaalaniz: 🤖 [Codex] Addressed in 82b39ea1 and refined in 0eb4ab23. The demo has a navigation bar control for the auto-scroll `animated` flag, and the button now shows `Animations: On` / `Animations: Off` so the current state is visible while recording the demo.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
aaalaniz: Sorry, AI missed here; let me explore addressing this workaround within the library itself.

aaalaniz: Actually, on second though, I do think this maneuver should be done within the client app.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -1468,52 +1577,142 @@ public final class ListView : UIView
...
+        return self.isUserScrollInProgress || self.scrollPositionInfo.isScrollInProgress
+    }
+
+    private func shouldDeferAutoScroll(_ info: _AutoScrollActionConfiguration) -> Bool {
+        guard info.scrollInterruptionPolicy == .deferDuringUserScrolling else {
+            return false
+        }
+
+        return self.isUserScrollInProgress || self.scrollPositionInfo.isScrollInProgress
+    }
```
johnnewman-square, line 1677: My agent had this final feedback:

> I think using Listable’s own isUserScrollInProgress is the cleaner fix for the deferral/skip logic. Right now the PR does this:
> 
> ```swift
> return self.isUserScrollInProgress || self.scrollPositionInfo.isScrollInProgress
> ```
> But `scrollPositionInfo.isScrollInProgress` includes `scrollView.isTracking`, so it can be true for a touch-down that never becomes a drag. That creates a pending auto-scroll even though Listable never entered its own “user scroll” lifecycle, so there may be no later delegate callback to flush it.
> 
> `isUserScrollInProgress` has the better lifecycle pairing:
> 
> - Set true in scrollViewWillBeginDragging.
> - Set false and flush in scrollViewDidEndDragging when there is no deceleration.
> - Set false and flush in scrollViewDidEndDecelerating when there is deceleration.
> 
> That means every deferred action caused by isUserScrollInProgress == true has a corresponding end callback that should flush it. Nice and tidy.

I have verified locally that touching down on the scroll view without dragging will enter the `isTracking` state, but will not call those delegates when the touch down gesture ends. I think this does create a small opportunity to enter a pending autoscroll. We could remove `|| self.scrollPositionInfo.isScrollInProgress` from both of these functions to omit the tracking case.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
johnnewman-square: This update is looking good!

ListableUI/Sources/ListView/ListView.swift
johnnewman-square: This is okay to me. Since this closure is stored, it would be good to add a small doc around this new closure to note that clients should only use weak self.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
johnnewman-square: The new behavior looks much smoother in the demo. And having this be a client app configuration tweak sounds good.

Development/Sources/Demos/Demo Screens/CustomAutoScrollingViewController.swift
johnnewman-square: Thank you for adding that flag -- it's nice to see both behaviors in the demo.

ListableUI/Sources/ListView/ListView.Delegate.swift
johnnewman-square: These updates look good 👍

johnnewman-square (approved):
This looks good. Thanks for making those edits. There were just a couple final items of feedback that I think we could address before merging.

ListableUI/Sources/ListView/ListView.swift
robmaceachern: My agent had the same finding.

- [x] Resolve before merging

robmaceachern (approved):
Thanks for doing this!

ListableUI/Sources/ListView/ListView.swift
aaalaniz: 🤖 [Codex] Addressed in 2a62c746. I added docs to the adjustment type/factory noting that declarative scrolling APIs may store the closure, and that clients should capture list owners/configurers weakly to avoid retain cycles.

ListableUI/Sources/ListView/ListView.swift
aaalaniz: 🤖 [Codex] Addressed in 370abbcd. The skip/defer policies now rely only on Listable’s own `isUserScrollInProgress` lifecycle flag, so a touch-down tracking state that never becomes a drag will not create a pending auto-scroll with no matching flush callback. I also added focused coverage for the non-dragging path.

PR #628, opened 2026-07-30
Add ScrollAnimation for duration-based scrolls

maxmeyers:
`UIScrollView` gives you a `Bool` for `setContentOffset(_:animated:)` and nothing else, so a programmatic Listable scroll is either instant or runs at a fixed system speed. This adds `ScrollAnimation` so a caller can say how long the scroll should take.

```swift
listActions.scrolling.scrollToSection(
    with: section.identifier,
    scrollPosition: .init(position: .top),
    animation: .duration(0.75)
)
```

The motivating case is a guided flow in Square Point of Sale: as the user completes each step, the list scrolls to the next one. At system speed the scroll is fast enough that it reads as a jump, and there's no way to slow it down.

### What's here

`ScrollAnimation` has three states — `.none`, `.system`, and `.duration(_:)` — and every `animated: Bool` scrolling method on `ListView` and `ListActions.Scrolling` now has an `animation:` counterpart. `animated: true` and `false` map to `.system` and `.none`, so existing callers behave exactly as before.

Two things came along with it, because the new path needed them and the old ones didn't have them:

- `scrollToTop(...)` and `scrollToLastItem(...)` now take a `completion` handler. They previously had no way to report finishing.
- Fixed a completion handler being stranded when `animated: true` was requested inside a `UIView.performWithoutAnimation` block. The suppressed scroll produced no `scrollViewDidEndScrollingAnimation(_:)` callback, so the handler queued for it was never called.

### Why a display link, and not `UIView.animate`

The usual suggestion for this problem is:

```swift
UIView.animate(withDuration: duration) {
    scrollView.setContentOffset(target, animated: false)
}
```

That animates, but on a `UICollectionView` **the scroll is completely blank**. The assignment lands the real `contentOffset` on the target immediately, so the collection view lays out at the destination and recycles every cell it was showing. Only the layer's `bounds` animate back — across a region that no longer has any cells in it. I had this built and passing tests before noticing it on screen; at ~100 rows and at ~800 rows of travel, content vanished for the whole animation and reappeared at the end.

Unit tests can't see it, which is worth knowing if this code is ever revisited: `layer.animationKeys()` contains `bounds.origin` and `layer.presentation()` interpolates correctly the entire time the screen is empty.

So `ScrollAnimationDriver` does what `UIScrollView` does for its own animation — assigns `contentOffset` on every frame from a `CADisplayLink`. Each assignment drives a layout pass, so cells dequeue as they scroll into view. `ListViewTests` has a regression test for this specifically (`test_scroll_with_duration_animation_keeps_content_laid_out`), asserting that the visible-item set mid-scroll is both non-empty and *different* from the set at the end — content laid out only at the destination would report the same items throughout.

The curve is a hardcoded ease-in-out, matching what `UIScrollView` does for its own scroll. It's a cosine rather than UIKit's bezier because a `CAMediaTimingFunction` only exposes its control points, not its value at a given time — solving the bezier per frame would mean carrying a Newton-iteration solver in Listable to land within 2% of where a one-line closed form already lands. There's no parameter for it: an exposed curve would have needed a bezier solver to be honest about the names, and no caller wanted one.

### Two details worth a reviewer's attention

**Where the animation is resolved.** A scroll toward content that hasn't been laid out yet is deferred until the presentation state catches up, and that deferred work runs on a later runloop pass — outside any `performWithoutAnimation` block the caller made the request from, where `UIView.areAnimationsEnabled` reads `true` again. So the animation is resolved against the current context where the scroll is *requested*, not where the offset changes. Resolving late would silently ignore a caller's request not to animate.

**Interruption.** A driven scroll is stopped by the same things that stop the scroll view's own animation — the user taking hold of the list (`scrollViewWillBeginDragging`), or another scroll superseding it — and reports its completion handler either way. It's also cancelled when the list's identifier changes, since the offset reset there invalidates the target, and in `ListView.deinit`, because the main runloop retains the display link and would otherwise keep driving a torn down list.

### Testing

Nine tests in `ListViewTests` under `// MARK: Scroll animations`, covering completion-exactly-once, offset parity with `.system`, mid-scroll sampling proving it eases rather than jumps, the blank-scroll regression above, cancellation, non-positive durations, suppressed animations, and that the driver doesn't retain itself while animating.

Worth noting for anyone extending these: a display-link animation is immune to `layer.speed`, so the `layer.speed = 4` that `show(vc:)` applies to the test host doesn't affect it. Durations in these tests are real wall-clock seconds.

Also verified by hand in the demo app. `ScrollToSectionCompletionHandlerViewController` and its new `ScrollToOffscreenSectionCompletionHandlerViewController` subclass (40 sections, so the far ones are nowhere near the initial layout) both have an animation picker with None / System / 1s / 3s.

### Checklist

### DIFF
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -1108,6 +1108,15 @@ public final class ListView : UIView
 
         switch animation.storage {
         case .none, .system:
+            // A scroll with nowhere to go starts no animation, so the scroll view never
+            // calls `scrollViewDidEndScrollingAnimation(_:)` and a handler queued for that
+            // callback would wait forever. Report it the way an unanimated scroll is
+            // reported instead.
+            guard collectionView.contentOffset != targetOffset else {
+                handleScrollCompletion(reason: .scrolled(animated: false), completion: completion)
+                return
+            }
+
             collectionView.setContentOffset(targetOffset, animated: animation.usesSystemAnimation)
             handleScrollCompletion(
                 reason: .scrolled(animated: animation.usesSystemAnimation),
@@ -1127,7 +1136,9 @@ public final class ListView : UIView
     /// `scrollRectToVisible(_:animated:)` compute where they land internally, so a driven
     /// animation has to perform the change unanimated and read the resulting offset back.
     ///
-    /// `changeOffset` receives whether it should ask the scroll view to animate.
+    /// `changeOffset` receives whether it should ask the scroll view to animate, and may be
+    /// invoked more than once — an unanimated call is how this discovers where the scroll
+    /// would land.
     ///
     /// Callers that have already determined the list cannot scroll should report
     /// `.cannotScroll` themselves instead of calling this.
@@ -1141,6 +1152,26 @@ public final class ListView : UIView
 
         switch animation.storage {
         case .none, .system:
+            // As in `applyScroll(with:to:completion:)`, an animated scroll that does not
+            // move reports nothing, stranding a queued handler. Only this variant cannot
+            // see that in advance, so it has to perform the change unanimated to find out.
+            // That costs a layout pass, so it is only worth doing when there is in fact a
+            // handler to strand.
+            if animation.usesSystemAnimation, completion != nil {
+                let startOffset = collectionView.contentOffset
+
+                changeOffset(false)
+
+                let movesContent = collectionView.contentOffset != startOffset
+
+                collectionView.setContentOffset(startOffset, animated: false)
+
+                guard movesContent else {
+                    handleScrollCompletion(reason: .scrolled(animated: false), completion: completion)
+                    return
+                }
+            }
+
             changeOffset(animation.usesSystemAnimation)
             handleScrollCompletion(
                 reason: .scrolled(animated: animation.usesSystemAnimation),
--- a/ListableUI/Sources/ListView/ListView.swift
+++ b/ListableUI/Sources/ListView/ListView.swift
@@ -137,9 +137,13 @@ public final class ListView : UIView
     }
     
     deinit
-    {        
+    {
         self.keyboardObserver.remove(delegate: self)
-        
+
+        // A driven scroll animation would otherwise keep running against a torn down list,
+        // because the display link driving it is retained by the main runloop.
+        self.cancelScrollAnimation()
+
         /**
          Even though these are zeroing weak references in UIKIt as of iOS 9.0,
          
@@ -561,6 +565,26 @@ public final class ListView : UIView
     }
 
     ///
+    /// Scrolls to the provided item, with the provided positioning, using the provided animation.
+    /// If the item is contained in the list, true is returned. If it is not, false is returned.
+    ///
+    @discardableResult
+    public func scrollTo(
+        item : AnyItem,
+        position : ScrollPosition,
+        animation : ScrollAnimation,
+        completion: ScrollCompletion? = nil
+    ) -> Bool
+    {
+        self.scrollTo(
+            item: item.anyIdentifier,
+            position: position,
+            animation: animation,
+            completion: completion
+        )
+    }
+
+    ///
     /// Scrolls to a custom vertical offset for the provided item.
     /// The adjustment receives the item's frame and visible content frame,
     /// then returns the vertical delta to apply.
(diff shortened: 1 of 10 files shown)


ListableUI/Sources/ListView/ListView.swift
```diff
@@ -914,6 +1093,129 @@ public final class ListView : UIView
...
+            changeOffset(false)
+
+            let targetOffset = collectionView.contentOffset
+
+            // Back to where the scroll started, so the driver can animate the distance
+            // itself. Restoring here rather than on the driver's first frame avoids
+            // displaying a frame at the target.
+            collectionView.setContentOffset(startOffset, animated: false)
+
+            driveScroll(to: targetOffset, duration: duration, completion: completion)
```
johnnewman-square, line 1193: Not a blocker at all, but one edge case occurred to me: if the cell heights dynamically change during the driven scroll animation, we may animate to the wrong end location. It might be difficult to add a test for this case and I'm happy to treat it as an edge case.

ListableUI/Sources/ListView/ListView.swift
```diff
@@ -914,6 +1093,129 @@ public final class ListView : UIView
...
+        to targetOffset: CGPoint,
+        completion: ScrollCompletion?
+    ) {
+        // Whatever this scroll is, it supersedes one that is still running.
+        cancelScrollAnimation()
+
+        switch animation.storage {
+        case .none, .system:
+            collectionView.setContentOffset(targetOffset, animated: animation.usesSystemAnimation)
+            handleScrollCompletion(
```
johnnewman-square, line 1121: 🤖 A `.system` scroll queues its completion even when the target offset equals the current offset. UIKit doesn’t start an animation or call `scrollViewDidEndScrollingAnimation(_:)` in that case, leaving the handler queued indefinitely. Could we detect no movement and complete asynchronously? This affects `scrollToLastItem` when already at the bottom and the corresponding `changeOffset` overload used by `scrollToTop` when already at the top.

johnnewman-square: - [ ] Investigate before merging.

johnnewman-square (approved):
Neat! This change looks good on my end and the demo works great. I left a couple comments: the one from my agent about the scroll completion handler feels like one worth investigating before we merge.

ListableUI/Sources/ListView/ListView.swift
maxmeyers: 🤖 Confirmed and fixed in e013df44 — good catch, and it's a regression this PR introduces rather than a latent one: `scrollToTop` and `scrollToLastItem` didn't take a completion handler before this change, so there was nothing to strand. A test reproducing both cases you named failed on the queued handler (`Exceeded timeout of 1 seconds, with unfulfilled expectations`) before the fix and completes in 0.014s after it. The explicit-offset `applyScroll` compares against the current offset directly; the `changeOffset` variant can't see the destination in advance, so it performs the change unanimated to discover it and then restores — the same trick the `.duration` path already used — and only pays that layout pass when a completion handler was actually supplied.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

maxmeyers: 🤖 Confirmed and fixed in e013df44 — good catch, and it's a regression this PR introduces rather than a latent one: `scrollToTop` and `scrollToLastItem` didn't take a completion handler before this change, so there was nothing to strand. A test reproducing both cases you named failed on the queued handler (`Exceeded timeout of 1 seconds, with unfulfilled expectations`) before the fix and completes in 0.014s after it. The explicit-offset `applyScroll` compares against the current offset directly; the `changeOffset` variant can't see the destination in advance, so it performs the change unanimated to discover it and then restores — the same trick the `.duration` path already used — and only pays that layout pass when a completion handler was actually supplied.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

ListableUI/Sources/ListView/ListView.swift
maxmeyers: 🤖 Agreed it's an edge case, and worth noting the system animation has it too: `setContentOffset(_:animated:)` commits to a target offset up front just as the driver does, so content that resizes mid-scroll can land off-target either way. The driven scroll is no worse than `.system` here rather than introducing a new failure mode, which is why I've left it. Clamping to the content size each frame would diverge from what the caller asked for, so if this ever needs solving I'd rather it be solved for both paths at once.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

johnnewman-square (approved):

